PR #1, opened 2026-04-04
Copy pasting public website frontend

sophp516:
I refactored our public website (except for the backend) so that it uses react router structure.
There is no backend call and everything is using mock data, so we can connect to dali-api once it's done.
I did add a login page just for now, which also does enable separate login for members, students, and partners.
Screenshot 2026-04-04 at 7 38 22 PM


henryscheible (approved):
LGTM

PR #5, opened 2026-04-05
Create Schema in Prisma

henryscheible:


dali-api/prisma/schema.prisma
```diff
@@ -28,5 +28,194 @@ model User {
...
+  applications         Application[]
+  domains              DomainApplicationCycle[]
+  challengeVersions    ChallengeVersionApplicationCycle[]
+}
+
+enum ApplicationStatus {
+  Draft
+  Submitted
+  Withdrawn
+}
```
sophp516, line 76: is there a reason we don't have other statuses like interview, rejected, etc?

dali-api/prisma/schema.prisma
sophp516: shouldnt we have interview scheduling related models as well?

dali-api/prisma/schema.prisma
```diff
@@ -28,5 +28,194 @@ model User {
...
+  userId        String
+  user          User        @relation(fields: [userId], references: [id])
 }
+
+model Application {
+  id        String   @id @default(cuid())
+  createdAt DateTime @default(now())
+  updatedAt DateTime @updatedAt
+
+  // answers: map of question UUID key -> answer value (JSON). Saved incrementally
```
sophp516, line 94: if answer is json, how do we deal with questions that involve uploading a file or something?

dali-api/prisma/schema.prisma
```diff
@@ -28,5 +28,194 @@ model User {
...
+// before they have created a User account.
+model DALIMember {
+  id        String   @id @default(cuid())
+  createdAt DateTime @default(now())
+  updatedAt DateTime @updatedAt
+
+  userId    String?  @unique
+  user      User?    @relation(fields: [userId], references: [id])
+
+  daliEmail String?  @unique
```
sophp516, line 50: fields necessary for hiring: hiredRoles (domain + level)
additionally, classYear, termsInDali, imageUrl, githubUsername?

dali-api/prisma/schema.prisma
```diff
@@ -28,5 +28,194 @@ model User {
...
+  domain       Domain    @relation(fields: [domainId], references: [id])
+  createdById  String
+  createdBy    User      @relation(fields: [createdById], references: [id])
+
+  domainApplications   DomainApplication[]
+  applicationCycles    ChallengeVersionApplicationCycle[]
+}
+
+enum ApplicationCycleStatus {
+  Draft
```
sophp516, line 199: if applicationCycleStatus is tied to the status of the application and applicationStatus is tied to the individual application, what does 'Draft' here refer to? Shouldn't it be 'Closed' or something similar?

sophp516 (changes requested):
Shouldn't we have interview scheduling related models as well?

dali-api/prisma/schema.prisma
```diff
@@ -28,5 +28,194 @@ model User {
...
+
+  daliEmail String?  @unique
+}
+
+model ApplicationCycle {
+  id        String   @id @default(cuid())
+  createdAt DateTime @default(now())
+  updatedAt DateTime @updatedAt
+
+  term      String
```
Kiran-Jones: should term be a plain string or enum/regex? might be confusing if we don't enforce a standard

dali-api/prisma/schema.prisma
Kiran-Jones, line 50: we already have daliEmail as a property of the user, do we want it in both schemas? not sure what the best design is curious what you think

dali-api/prisma/schema.prisma
```diff
@@ -28,5 +28,194 @@ model User {
...
+  applicationCycle         ApplicationCycle       @relation(fields: [applicationCycleId], references: [id])
+  // Pinned at application creation time so that form updates do not affect in-progress drafts.
+  applicationFormVersionId String
+  applicationFormVersion   ApplicationFormVersion @relation(fields: [applicationFormVersionId], references: [id])
+
+  statusUpdates            ApplicationStatusUpdate[]
+  domainApplications       DomainApplication[]
+}
+
+model DomainApplication {
```
Kiran-Jones, line 114: should this also have a domain field? the list of DomainApplication in Application might be confusing with out it

dali-api/prisma/schema.prisma
Kiran-Jones: key question value is location of file?

sophp516: I guess that works together with question.type

dali-api/prisma/schema.prisma
sophp516: yeah we should prob have a domain enum

dali-api/prisma/schema.prisma
henryscheible: I think hired roles will be a separate model (though I guess it will have a back relation). I think we can skip the other ones for now but I agree we'll need them later

henryscheible: I was thinking both so that when a new member signed in with their DALI email that would allow their DALIMember to be matched to their User, but I guess its redundant outside of that. I feel like leave for now and if we want to get rid of it later it should be pretty easy

dali-api/prisma/schema.prisma
henryscheible: Hmmm. Honestly maybe we could delete term. We will have dates that it opens, closes, and decisions are released in ApplicationCycleStatusUpdates so if we wanted to sort by date we could use those, and if we want to include term in the name we could put it in the name of the cycle. Should add a name property here I guess. I think it is nice because we might have cycles that are weird (like only one domain) and naming them whatever we want gives us that flexibility.

dali-api/prisma/schema.prisma
henryscheible: Those will probably be in a separate decisions table (to keep track of which DALI member made the decision, whether the decision has been communicated to the applicant, etc). The distinction in my head is whether the status change is applicant or dali-initiated. These are the applicant-initiated ones

dali-api/prisma/schema.prisma
henryscheible: Yeah agree with Kiran. Maybe we have an attachments table that it points to, but I feel like we can leave that as an implementation detail for v1. Actually thinking we implement all this with textual answer questions only, then go back and add the other kinds.

dali-api/prisma/schema.prisma
henryscheible: Should be computable as ChallengeVersion.domain. There will never be a DomainApplication that isn't solving a particular challenge

dali-api/prisma/schema.prisma
henryscheible: Draft means that the hiring lead and domain leads are writing the questions for the application but it isn't open to being filled out yet. Open means that people can submit, UnderReview means from the due date to when we have told everyone their decision, and Decisions released is all time after that

sophp516 (approved):

PR #10, opened 2026-04-07
Sso

Kiran-Jones:
- added dartmouth and google SSO backend routes
- created two model schemas: `OAuthSession` and `RefreshToken`
- added simple account merging (dartmouth sign in/google sign in) logic
- added one get self route

- added unit tests
- added npm run test ci on push and pr


sophp516 (approved):
LGTM

PR #29, opened 2026-04-10
Fix prisma migrate deploy: add url to schema datasource block

henryscheible:
## Summary
- `prisma.config.ts` is not copied into the Docker image (only `./prisma` is), so the release machine had no `DATABASE_URL` when running `prisma migrate deploy`
- Adds `url = env("DATABASE_URL")` directly to the `datasource db` block in `schema.prisma` as the authoritative source

## Test plan
- [ ] Merge and verify deploy succeeds — release command should complete without the "datasource.url property is required" error
- [ ] `GET /admin` returns 200 (no more missing column errors)


Kiran-Jones (approved):
LGTM

PR #34, opened 2026-04-12
Add Google OAuth login to dali-api

henryscheible:
## Summary
- Adds a `/login` page to `dali-api` with a "Sign in with Google" button, restricted to `@dali.dartmouth.edu` accounts
- Implements server-side OAuth callback at `/auth/callback/google` — no PKCE needed since both endpoints are on the same server; CSRF protection via a short-lived httpOnly state cookie
- Adds auth guard to the layout route, redirecting unauthenticated users to `/login`
- Replaces hardcoded `admin@dali.dartmouth.edu` user lookup in the admin action with the real session user
- Removes mock user from Layout header; displays the actual logged-in user's name and email

## Test plan
- [ ] Visit `/admin` with no cookies → redirected to `/login`
- [ ] Click "Sign in with Google" → Google OAuth, accepts only `@dali.dartmouth.edu`
- [ ] After auth → redirected to `/admin`, header shows real name and email
- [ ] Creating a hiring cycle attributes it to the logged-in user
- [ ] Non-DALI Google account → rejected, redirected to `/login?error=access_denied`

### DIFF
--- a/dali-api/app/components/Layout.tsx
+++ b/dali-api/app/components/Layout.tsx
@@ -7,12 +7,12 @@ import {
   Trophy,
   ClipboardList,
 } from 'lucide-react'
-import { currentUser, adminUser } from '~/mockData'
 import type { ViewMode } from '~/types'
 interface LayoutProps {
   children: React.ReactNode
   viewMode: ViewMode
   setViewMode: (val: ViewMode) => void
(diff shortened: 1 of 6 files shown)


Kiran-Jones (approved):
LGTM

PR #38, opened 2026-04-12
Admin / mentor side application review

sophp516:


dali-api/prisma/migrations/20260412201234_add_mentor_review/migration.sql
henryscheible, line 1: Why do you edit this? Shouldn't you create a new migration on top of it? I think that is why the migration safety check is failing

henryscheible (dismissed):
The migration change should be the addition of a migration I think

dali-api/prisma/migrations/20260412201234_add_mentor_review/migration.sql
henryscheible: Oop looks like it succeeded now, but I still think best practice would be to not edit the migration? I might be wrong. I think the dev deployment will fail if we push it like this (even though it works locally and the ci checks think that it is fine)

sophp516: it only adds the MentorReview table and nothing else was edited. would that still cause issues?

henryscheible: Oops I got super confused. I think you are good!

Kiran-Jones (approved):
LGTM

PR #39, opened 2026-04-12
Fail CI when migration files are deleted in a PR

henryscheible:
## Summary

- Adds a `git diff` check to `migration-check.yml` that fails if any files under `prisma/migrations/` are deleted in a PR
- Closes the gap where deleting and replacing a migration with the same net effect would silently pass the existing `prisma migrate diff` check

## Test plan

- [ ] Open a PR that deletes a migration file and confirm the check fails
- [ ] Confirm normal PRs (no deleted migrations) still pass

### DIFF
--- a/.github/workflows/migration-check.yml
+++ b/.github/workflows/migration-check.yml
@@ -25,6 +25,8 @@ jobs:
 
     steps:
       - uses: actions/checkout@v4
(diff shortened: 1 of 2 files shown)


Kiran-Jones (approved):
LGTM

PR #50, opened 2026-04-14
Enforce single active cycle and anchor mentor view on it

ashnaghanate:
## Summary

Right now nothing prevents two hiring cycles from being in active states (Open or Closed) simultaneously, and the mentor dashboard just picks "the most recent cycle the mentor has any reviewer assignment on" — which gets confusing once a second cycle is set up. This PR introduces a clear single-active-cycle invariant and rewires the mentor loader around it.

## What "active" means

A cycle is **active** when its latest status is **Open** or **Closed** — these are the stages where applicants are submitting and reviewers are reading/interviewing. **Draft** and **DecisionsReleased** are not active. The invariant: **at most one cycle can be active at a time.**

### `app/lib/cycles.ts` (new)
Two helpers:
- **`getActiveCycle()`** — returns the single currently-active cycle (latest status in {Open, Closed}) along with its `currentStatus`, or `null`.
- **`findOtherActiveCycleId(excludingCycleId)`** — returns the id of any *other* active cycle, used to enforce the invariant on transitions.

### `api.cycles.$cycleId.status.ts` — transition guard
When advancing a cycle into **Open** or **Closed**, the action now checks whether any *other* cycle is currently active. If so it returns **409 Conflict** with a structured error:

```json
{
  "error": "Another cycle is already active. Only one cycle can be Open or Closed at a time. Move the existing active cycle to DecisionsReleased first.",
  "activeCycleId": "cycle-fall-2026"
}
```

Open → Closed within the **same** cycle is still allowed (only one cycle is involved). DecisionsReleased frees the slot for the next cycle.

### `mentor.tsx` loader — anchor on active cycle
Previously: `prisma.cycleReviewer.findFirst({ where: { daliMemberId }, orderBy: applicationCycle.createdAt desc })` — returned the latest cycle the mentor had any assignment on, regardless of whether that cycle was actually running.

Now:
1. Call `getActiveCycle()`. If null → empty state.
2. Look up the mentor's `cycleReviewer` row scoped to **that specific cycle** via the `daliMemberId_applicationCycleId` unique index. If none → empty state (mentor has no role in the active cycle).
3. Use the active cycle's `currentStatus` for the stage mapping.

Side effect: I dropped `Draft` and `DecisionsReleased` from `cycleStatusToStage` since those statuses can no longer reach this loader.

### Seed update
The previous seed left **Winter 2028 in Open** while **Fall 2026 was Closed** — that's two active cycles, violating the new invariant. Winter 2028 now stays in **Draft** in the seed; an admin would advance it to Open after Fall 2026 reaches DecisionsReleased. Log line updated to highlight which cycle is active.

## Verification

End-to-end against the dev server, all four scenarios behave correctly:

```
─── 1. Riley (Eng reviewer, Fall 2026) fetches mentor view ───
"activeCycle": cycle-fall-2026, "currentStage": "interviews"  ✓

─── 2. Try to advance Winter 2028 (Draft) → Open while Fall 2026 is Closed ───
{"error":"Another cycle is already active...","activeCycleId":"cycle-fall-2026"}
HTTP 409  ✓

─── 3. Advance Fall 2026 Closed → DecisionsReleased ───
{"currentStatus":"DecisionsReleased"}  ✓

─── 4. Now advance Winter 2028 → Open ───
{"currentStatus":"Open"}  ✓

─── 5. Riley (no reviewer assignment on Winter 2028) re-fetches mentor view ───
"activeCycle": null, "currentStage": "challengeSetup", "allCycleApps": []  ✓
```

- `npm run typecheck` — clean for files I touched
- `npm test` — 58/58 pass

## Tradeoff (also discussed when scoping)

This is **API-layer enforcement**, not a database constraint. Two admins racing to open cycles in the same second could both succeed since there's no transactional lock around the read-then-write. For a hiring tool with a small admin pool that's almost certainly fine; if it ever becomes a problem the path is a partial unique index on a derived `isActive` column.

## Test plan

- [ ] `docker compose down -v && docker compose up` — fresh DB
- [ ] Visit `http://localhost:3001/dev-login-as?netId=rev001&redirect=http://localhost:3001/mentor` — mentor dashboard shows Fall 2026 in interview stage
- [ ] Admin: try to advance Winter 2028 to Open via `/admin/cycle/cycle-winter-2028` — should see error toast / 409
- [ ] Advance Fall 2026 → DecisionsReleased, then advance Winter 2028 → Open — succeeds
- [ ] Reload mentor view — Riley (Fall 2026 reviewer only) sees the empty state because the active cycle changed

### DIFF
--- a/dali-api/app/lib/cycles.ts
+++ b/dali-api/app/lib/cycles.ts
@@ -1,0 +1,44 @@
+import { prisma } from "~/lib/db";
+
+// A cycle is "active" when its latest status is Open or Closed — those are the
+// stages where applicants are submitting and reviewers are reading/interviewing.
+// Draft and DecisionsReleased are not active. By convention we enforce at most
+// one active cycle at a time (see api.cycles.$cycleId.status.ts).
+export const ACTIVE_STATUSES = ["Open", "Closed"] as const;
+export type ActiveStatus = (typeof ACTIVE_STATUSES)[number];
+
+/**
+ * Find the single currently-active cycle, or null if none.
+ *
+ * If the invariant is somehow violated (multiple cycles active at once),
+ * returns the most recently created one and the caller can ignore the rest.
+ */
+export async function getActiveCycle() {
+  const cycles = await prisma.applicationCycle.findMany({
+    include: {
+      statusUpdates: { orderBy: { createdAt: "desc" }, take: 1 },
+    },
+    orderBy: { createdAt: "desc" },
+  });
+
+  for (const cycle of cycles) {
+    const latest = cycle.statusUpdates[0]?.newStatus;
+    if (latest && (ACTIVE_STATUSES as readonly string[]).includes(latest)) {
+      return { ...cycle, currentStatus: latest as ActiveStatus };
+    }
+  }
+  return null;
+}
+
+/**
+ * Returns the id of any *other* cycle (i.e. not `excludingCycleId`) that is
+ * currently active, or null if none. Used to enforce single-active-cycle when
+ * advancing a cycle into Open.
+ */
+export async function findOtherActiveCycleId(
+  excludingCycleId: string,
+): Promise<string | null> {
+  const active = await getActiveCycle();
+  if (!active || active.id === excludingCycleId) return null;
+  return active.id;
+}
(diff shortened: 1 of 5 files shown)


Kiran-Jones (approved):
LGTM

PR #55, opened 2026-04-14
Cycle management redesign spec

henryscheible:
## Summary

- Documents the full redesign of the application cycle system around a four-phase lifecycle: **Draft → Open → Under Review → Completed**
- Covers all planned schema changes (12 sections): new models for `Decision`, `ApplicationReview`, `DelibsSession`, `CycleInterviewer`, `InterviewNoteVersion`; renamed enums; moved/nullable fields
- Specifies per-`DomainApplication` status inference (8 statuses, never stored — always derived)
- Phase-by-phase feature detail including reviewer assignment, delibs kanban flow, interview scheduling with in-domain + cross-domain interviewers, and versioned interviewer notes
- Key query patterns and concurrency/race condition plans

This is a spec-only PR for team review before implementation begins.


CYCLE_REDESIGN_SPEC.md
```diff
@@ -0,0 +1,767 @@
...
+  id        String   @id @default(cuid())
+  createdAt DateTime @default(now())
+
+  domainApplicationId String
+  domainApplication   DomainApplication @relation(fields: [domainApplicationId], references: [id])
+
+  type  DecisionType
+  stage DecisionStage
+
+  madeById String
```
Kiran-Jones, line 156: what's the rationale for having this? would it always be the domain lead/whoever is running delibs v2?

CYCLE_REDESIGN_SPEC.md
```diff
@@ -0,0 +1,767 @@
...
+|---|---|
+| `Closed` | `UnderReview` |
+| `DecisionsReleased` | `Completed` |
+
+---
+
+### 2. Add `closeDate` to `ApplicationCycle`
+
+```prisma
+model ApplicationCycle {
```
Kiran-Jones, line 23: General comment (not necessarily specific to this model) but should we also add support for other cycle reviewers but not cycle reviewers assigned to a given app to provide comments? Notion currently does this (comment section) and people have done this in the past/left comments mid delibs. Not ultra important but nice to have

CYCLE_REDESIGN_SPEC.md
```diff
@@ -0,0 +1,767 @@
...
+
+**Submitting** = `POST /api/reviews/:id/submit` → sets `submittedAt = now()` and `submittedById = auth.user.memberId`. Read-only while submitted.
+
+**Unsubmitting** = `POST /api/reviews/:id/unsubmit` → clears `submittedAt` and `submittedById`. Review becomes editable again. The reviewer can then re-submit, which re-stamps both fields.
+
+**Changing assignment** (before submission): delete the blank record and create a new one for a different `cycleReviewerId`.
+
+---
+
+### 10. New `DelibsSession` model
```
Kiran-Jones, line 212: Views of this enforced in application level, should make it so that being a application reviewer OR interviewer is view criteria. This will technically allow reviewers to see interview board but that doesn't really matter. If we care about this, could add property to DelibsSession model  being applicationReview or interviewReview (for example) and filter by those too

Kiran-Jones (commented):
Do we want

CYCLE_REDESIGN_SPEC.md
ashnaghanate: it does seem good to have just bc we can have multiple domain leads and sometimes r forgetful

CYCLE_REDESIGN_SPEC.md
ashnaghanate: probably actually want all reviewers to be able to the see the interview board because we sometimes still invite them to delibs

ashnaghanate (approved):
Looks great!

CYCLE_REDESIGN_SPEC.md
```diff
@@ -0,0 +1,767 @@
...
+```
+
+`Interview.status = Completed` is set when either interviewer submits the recommendation via "Mark Complete". The recommendation field is required to mark complete. The other interviewer can still add/edit their own notes after completion.
+
+---
+
+### 7. New `InterviewNoteVersion` model
+
+Per-interviewer notes are versioned — each auto-save creates a new append-only record. Current notes = latest record by `createdAt`. Interviewers write only their own notes but can read each other's (both are visible during the interview).
+
```
sophp516, line 105: do we record the interview questions asked? might be somewhat helpful for gauging domain specific questions and their difficulty

PR #56, opened 2026-04-15
Gmail and s3

sophp516:
## Summary
- Adds Gmail OAuth2 integration for sending transactional emails from applications@dali.dartmouth.edu (application received + status update templates)
- Adds S3 presigned URL upload flow for file attachments
- Adds API routes: api.email.send, api.upload.presign, api.upload.url, admin.authorize-gmail, admin.authorize-gmail.callback
- Updates api.my-application and ApplicantPortal to support new upload/email flows


Kiran-Jones (commented):
For all of the templates it would be nice to have them editable not in code (admin/hiring lead view). Contents look good but we could store them in postgres - this would also allow us to track version history of templates

henryscheible:
Would also be nice to think about integrating this into the schema a bit more... Specifically, instead of having the 6 hardcoded emails, what if the emails were parameterized by DecisionType. Then we can have these more closely linked to Decisions with less custom code. Like "Releasing" a decision (something already implemented for the hiring lead) should be sending an email, so it could just look up the email template with the DecisionType associated to that decision. It is totally fine to add more decision types (distinguishing rejection before/after interview) if necessary.

Kiran-Jones (approved):
LGTM merging, comments made into GH issues

PR #62, opened 2026-04-16
Three-environment deploy pipeline (dev/staging/prod)

henryscheible:
## Summary

- **Dev** (`dev` branch): Wipes the Neon dev database on every deploy, applies all Prisma migrations from scratch, then seeds. Fixes the current deploy failure caused by enum conflicts with prod data.
- **Staging** (`staging` branch): Restores the Neon staging database from prod, then applies new migrations. Catches data-incompatible migrations before they reach prod.
- **Prod** (`prod` branch): Unchanged — just applies migrations.

### Other changes
- Dockerfile now accepts an `OMIT_DEV` build arg so dev gets full dependencies (needed for `tsx` to run seeds)
- Removed the partial unique index from `seed.ts` (app-level transaction isolation is sufficient)
- Removed the `deploy-web` job from the workflow
- Updated `dali-api/README.md` deployment docs

### Prerequisites before merging
- [ ] Create a Neon `staging` branch in the project
- [ ] Create Fly app `dali-api-staging` with secrets (DATABASE_URL, etc.)
- [ ] Create a `staging` git branch

## Test plan
- [ ] Push to `dev` and confirm the Fly Deploy workflow succeeds (all migrations from scratch + seed)
- [ ] Push to `staging` and confirm the staging DB restores from prod and new migrations apply
- [ ] Push to `prod` and confirm only migrations run with no DB prep step

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #65, opened 2026-04-16
Fix dev deploy: wipe Neon dev DB via psql, not bogus /sql endpoint

henryscheible:
## Summary
- The "Wipe dev database" step added in #62 calls `POST https://console.neon.tech/api/v2/projects/{id}/branches/{id}/sql` — an endpoint that does not exist (verified against Neon's OpenAPI spec at https://neon.com/api_spec/release/v2.json). `curl -sf` turns the 4xx into exit code 22, killing every dev deploy before `flyctl deploy` runs.
- Replace the call with a `psql` invocation that runs `DROP SCHEMA public CASCADE; CREATE SCHEMA public;` directly against the Neon dev Postgres via a new `DEV_DATABASE_URL` repo secret.
- Staging's `/restore` step is a real, documented endpoint and is left untouched.

## Required before merging
- [ ] Add repo secret `DEV_DATABASE_URL` — set to the Postgres connection string for the Neon `development` branch (same value Fly already has as `DATABASE_URL` on `dali-api-dev`). Without this secret set, the workflow will still fail.

## Test plan
- [ ] After the secret is in place, merge and watch the `Fly Deploy` run on `dev`
- [ ] Confirm the "Wipe dev database" step succeeds
- [ ] Confirm `deploy-api` runs and the Fly release_command (`prisma migrate deploy && prisma db seed`) completes — `fly logs -a dali-api-dev`
- [ ] Spot-check that dev has a freshly seeded schema

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -16,17 +16,10 @@ jobs:
       - name: Wipe dev database
         if: github.ref_name == 'dev'
         env:
-          NEON_API_KEY: ${{ secrets.NEON_API_KEY }}
-          PROJECT_ID: falling-morning-15858013
+          DEV_DATABASE_URL: ${{ secrets.DEV_DATABASE_URL }}
         run: |
-          BRANCHES=$(curl -sf "https://console.neon.tech/api/v2/projects/$PROJECT_ID/branches" \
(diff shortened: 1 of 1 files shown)


ashnaghanate (approved):
looks great Henry!

PR #71, opened 2026-04-17
Applicant portal, admin UX redesign, and pipeline improvements

ashnaghanate:
## Summary

Comprehensive update: applicant-facing portal, admin/domain-lead UX overhaul, and pipeline blocker messaging.

### Applicant Portal (new)
- `/portal` — multi-stage portal (ApplicationOpen → Pending → InvitedToInterview → InterviewScheduled → PostInterviewPending → Accepted/Rejected/Waitlisted)
- `/portal/apply` — application form with real questions from ChallengeVersion, draft creation, auto-save, validation
- Unified `/login` with DALI Member (Google) and Dartmouth Student (CAS) auth paths
- `/auth/callback/cas` for CAS ticket validation
- Falls back to user's most recent application when cycle is Completed
- Interview booking per SCHEDULING.md spec

### Navigation Redesign
- 4 grouped tabs: Domain, Reviews, Cycles, Members
- Inline sub-pills (Reviews → Applications/Interviews, Cycles → Overview/Challenges/Rubrics)
- Dark DALI-branded header with coral accents and section-bg page background

### Domain Lead Dashboard
- Collapsible sections with progress badges
- Single active cycle only
- Setup section with challenge/rubric quick links in all phases (not just Draft)
- Full-context application detail: two-column layout with reviewer scores, interview info, decision history
- Interview tracking: "Awaiting Booking" list + booked interviews table

### Pipeline Blocker Messages
- Domain lead: "Set rubric before marking ready", "Waiting on hiring lead for general rubric", auto-assign checks reviewer count, interviewer availability warning
- Reviewer: stage progress indicator (Reviews → Availability → Interviews → Decisions), finalDelibs stage now reachable
- Interviewer: availability warning banner
- Applicant: improved empty-slots message

### Cycle Management
- Active cycle hero card + collapsed past cycles
- General form/rubric pickers with save/change/locked states
- Draft checklist requires general form + rubric
- Completion modal shows blockers, requires explicit force-close
- Auto-close cycles past their close date

### Submission URL Checking
- `lib/submission-check.ts` + `api.check-url.ts` validate GitHub/Figma URLs at submit time
- Wired into ChallengeBuilder

### Auth / Access Fixes
- /api/members, /api/reviewers, /admin-console accessible to domain leads and hiring leads
- Challenge/rubric routes allow domain leads
- dev-login routes applicants to /portal
- Delibs close creates Final decisions (not Draft)
- Collab server handles EADDRINUSE gracefully

### Bug Fixes
- Final delibs Prisma query: `interview` → `interviews: { some: ... }`
- Domain lead loader fallback includes all required fields
- Portal Date serialization in Google Calendar URL
- Vite SSR externals for collab packages

## Test plan
- [ ] Applicant portal: log in as different seed applicants, verify each stage
- [ ] Domain lead: collapsible sections, setup links, blocker messages
- [ ] Reviewer: stage progress indicator, availability/interview stages
- [ ] Cycle management: form/rubric pickers, checklist, completion modal
- [ ] Nav: 4 tabs with sub-pills, correct active states
- [ ] Auth: member → /reviewer, applicant → /portal


ashnaghanate:
Merged into #72 which now contains all changes (applicant portal + admin UX + pipeline improvements).

henryscheible (changes requested):
### Admin UX review

**1. Nav: Challenges and Rubrics nested under Cycles**

`Layout.tsx:51-61`. Challenges and Rubrics are nested as sub-pills under the **Cycles** tab, but in the schema they're independent entities — `Challenge` (`schema.prisma:163`) and `Rubric` (`schema.prisma:524`) own their own lifecycle and are only linked to cycles through `ChallengeVersionApplicationCycle` (`schema.prisma:251`). Nesting them under Cycles implies they're cycle-scoped, which can mislead users about how to edit or reuse them across cycles. Consider promoting them to top-level tabs or grouping them elsewhere (e.g., a "Content" section alongside Members).

**2. "Edit →" on Domain Challenge is misleading**

`domain-lead.tsx:550-552`. When the cycle is "Configured" and reviewers have been assigned, the Domain Rubric row correctly shows `(locked — reviewers have been assigned)`, but the Domain Challenge row still offers a blue "Edit →" link. The link routes to `/challenges/{id}`, which only supports *appending* a new `ChallengeVersion` via the `create-version` action (`admin.challenges.$id.tsx:42-58`) — it doesn't (and can't) edit the version the applicants actually answered. Users clicking "Edit" expecting to fix a typo will either create an unused new version or be confused.

Options:
- Mirror the rubric's locked state: show `(locked — reviewers have been assigned)` and hide the Edit link.
- Rename "Edit →" to "Manage versions →" or "View →" so it doesn't promise in-place editing.
- Link directly to the pinned ChallengeVersion in read-only mode.

**3. Duplicate hero-cycle selection logic**

`admin.tsx:137-143` and `admin.tsx:189-195`. `ActiveCycleHero` computes which cycle to feature at the top, and `PastCycles` re-computes the same expression to exclude it from the past-cycles list. If someone later changes the hero selection rule (e.g. prefer `UnderReview` over `Open`, or ignore stale Drafts) and only updates one site, the active cycle will either be shown twice or hidden from both lists.

Extract into a shared helper and call it from both components.

ashnaghanate:
Reopening — this PR contains everything. Closing #72 instead.

henryscheible (changes requested):
### Admin UX review (round 2)

**Rubric incorrectly required to mark challenge ready (spec violation)**

`domain-lead.tsx:860-876` (commit `08980a8`). The "Ready to finalize?" card blocks the "Mark Configuration as Ready" button when no rubric is set (`disabled={!currentRubricVersionId}`) and shows a yellow "Set a domain rubric below before marking as ready" message.

CYCLE_REDESIGN_SPEC.md:388-396 explicitly lists rubric assignment under **"Optional in Draft (can continue into Open)"**, not in the enforced Open checklist. Line 417 further notes rubric is *still mutable in Open*. Requiring it to mark-ready is a spec violation and prevents the flexibility the spec intends (e.g., opening applications while still iterating on the rubric).

Remove the `!currentRubricVersionId` gate and the yellow blocker message.

**Rubric belongs with reviewers, not challenge setup**

Related: the "Setup" section currently mixes applicant-facing config (challenge questions) with reviewer-facing config (rubric). Rubrics are never visible to applicants — they're tools reviewers use to score. Consider moving rubric selection into the "Team" section (renaming it "Review Setup" or "Reviewers" if it feels off) so it lives alongside the reviewers who'll use it, while "Setup" stays focused on what applicants see.

henryscheible (changes requested):
### Applicant portal + misc review (round 3)

**1. Role selection is locked after starting the application**

`portal.apply.tsx:578-625` (domain picker) only renders in the pre-draft state. Once `create-draft` runs, the applicant sees their selected domains as read-only pills (`:640-653`) and the action has no `add-domain` / `remove-domain` intents. An applicant who decides mid-way that they want to add or drop a domain has no UI path — and since there's no draft-delete either, they're stuck with their original selection through submission.

Suggested:
- Allow editing `selectedDomainIds` from the application form itself (the same picker UI, shown inline instead of gated behind `!draft`).
- On add: create a new `DomainApplication` for the added domain with empty answers.
- On remove: delete the corresponding `DomainApplication` (or soft-delete). Warn before destructive removal.
- Ownership check on the mutation (same gap as the `save-draft` / `submit` actions — needs `application.userId === auth.user.sub`).

**2. Submit button gives no feedback when submission is soft-blocked**

`portal.apply.tsx:491-574`, `:636`, `:719-727`. When a submit attempt hits URL warnings, the server returns `{ urlWarnings }` without redirecting (`:240-242`), leaving the application in Draft. The client:
- Renders a yellow warning banner inline below the questions (can be below the fold).
- Silently swaps the primary CTA from "Submit Application" → "Submit Anyway" (amber vs. coral — visually similar).
- Leaves the hardcoded "Draft" badge (`:636`) in place — a static label that doesn't react to submit state at all.

The applicant's most likely read is "I clicked Submit and nothing changed." Suggested: live status badge (Draft / Submitting / Action required / Submitted), scroll the warning into view on state change, and a more obvious button transition (modal confirmation, clearer copy).

**3. URL check indicator can render as a bare icon with no text**

`portal.apply.tsx:266-292`. `UrlCheckIndicator` renders `state.result.message` unconditionally in the amber branch, but `state.result` can now come back as `{ error: "Too many requests" }` (rate limit, 429) or `{ error: "Request body too large" }` (413) — shapes with no `message` field. Result: icon shown, text empty, applicant can't tell what's wrong. Fix by guarding on `message` presence and falling back to a generic string, or by normalizing the response shape in `api.check-url.ts` so all error paths include `{ status, message }`.

**4. URL-check rate limit is too aggressive for the submit flow**

`api.check-url.ts:7-8` sets max 5 / 60s. Every URL field calls `/api/check-url` on blur. A normal applicant filling in one GitHub and one Figma link, pasting/fixing typos, can trip this easily — and there's no client-side throttle. Raise the limit, key it per-user rather than per-IP, or skip the client-side blur check and rely on the server-side submit-time check alone.

**5. Figma `/design/` URLs are falsely flagged as private**

`lib/submission-check.ts:132`. The check hardcodes `https://www.figma.com/file/${fileKey}` regardless of whether the input URL was `/file/` or `/design/`. Figma redirects `/file/{key}` → `/design/{key}` for any file created in the current UI, and the checker treats any 3xx as "Figma file is private (redirects to login)" (`:137-143`). Result: every `/design/` URL (which is most of them today) false-positives. Fix by either preserving the original path segment or by accepting a redirect to the same hostname as OK.

**6. "Add Reviewer" throws at runtime — action imports don't include `isDomainLead`**

`api.cycles.$cycleId.reviewers.ts:4,34`. The action's auth check calls `isDomainLead(...)` but the import only brings in `isHiringLead` and `hasCycleAccess`. Every POST throws `ReferenceError: isDomainLead is not defined` → 500 response → client's `addReviewer` silently no-ops. Domain leads can't add reviewers to their own cycles. Fix: add `isDomainLead` to the import. The sibling file `api.cycles.$cycleId.reviewers.$reviewerId.ts` (DELETE path) likely has the same issue — please double-check.

**7. Reviewer and interviewer dashboards shouldn't be merged**

`Layout.tsx:40-49` nests the reviewer (`/reviewer`) and interviewer (`/interviewer`) surfaces as sub-pills under a single top-level "Reviews" tab. That framing is misleading:

- **They're different roles played by different people.** A cycle reviewer and a cycle interviewer need not be the same member — `CycleReviewer` and `CycleInterviewer` are independent relations in the schema. The sub-pill layout implies one view is a refinement of the other for a shared role.
- **It forces interviewers to wait out the review phase before they can provide availability.** `mentor.tsx:231` gates the availability form on `currentStage === 'collectingAvailability'`, and `mentor.tsx:43-49` only promotes the cycle to that stage once Released `InvitedToInterview` decisions exist. A member who is only an interviewer (not a reviewer) has nothing to do on `/reviewer` during `readingApplications` and can't submit availability until reviews wrap. In practice they should be able to mark availability as soon as they're added as a `CycleInterviewer`.

Split the two back into top-level tabs (or at least decouple the availability form from the cycle-stage gate so interviewers can enter availability at any time after invitation).

**8. Close delibs should create Draft decisions, not Final**

`api.delibs.$id.ts:89-102` (changed in this PR). Closing a delibs session now creates `Decision` rows at `stage: "Final"`. This skips the Draft stage and removes the domain lead's ability to revisit/adjust decisions after the delibs session closes — the only correction path is to create a new superseding Decision record.

Per `schema.prisma:425-429`, the stage sequence is `Draft → Final → Released` for a reason: Draft is where the domain lead can iterate, Final locks the decision for the hiring lead to release. Collapsing that straight to Final removes the review/adjustment step.

Revert to creating `Draft` decisions on close. The hiring lead (or domain lead) can explicitly advance them to Final as a separate action when they're confident.

**9. Progress ticker and portal collapse multi-domain applications into a single "primary" state**

`portal.tsx:752-772`. When an applicant has multiple domain applications in different states, the portal picks one "primary" stage using this priority order:

```
Accepted > InterviewScheduled > PostInterviewPending > InvitedToInterview
  > Waitlisted > Pending > Rejected > ApplicationOpen
```

Concrete consequences:
- **Accepted (Design) + Pending (Engineering)** → ticker shows "Decision"; the still-in-progress Engineering application is invisible.
- **Rejected (Design) + Pending (Engineering)** → ticker shows "Review"; the Rejected outcome is hidden until Engineering resolves.
- **Waitlisted (Design) + Pending (Engineering)** → ticker shows "Decision".

The priority is also inconsistent about negative outcomes: `Waitlisted > Pending > Rejected` means partial-success surfaces but partial-failure gets buried under any still-pending domain. The same primaryStage/primaryDa pair drives the main content card below the ticker, so the applicant never sees more than one domain at a time.

**Recommendation:** redesign the portal to render each domain application's status independently rather than trying to merge them. Drop the progress ticker as a top-level element (or render one per domain), and let each domain application surface its own stage view (interview booking, accepted/rejected, etc.) in the main content area. Applicants who applied to multiple domains actually need to track them all.

henryscheible (dismissed):
### Review (round 4) — additional findings from a deeper pass

**1. Waitlist rank off-by-N** (blocker)

`api.delibs.$id.ts:79-100` (Final delibs close). The outer `decisions.entries()` loop walks Accepts first, then Waitlist, then Rejects. `waitlistRank: d.type === "Waitlisted" ? index + 1 : null` uses the outer index, so the first Waitlisted applicant gets rank `N_accepts + 1` instead of `1`. The inner destructure `for (const [rank, id] of (columnOrder["Waitlist"] ?? []).entries())` captures the correct rank from the Waitlist array, but then throws it away. Use that inner `rank` value (or a Waitlist-local counter) instead of the outer `index`. As-is, no waitlist rank 1 ever exists when any Accept decisions are present in the session — downstream logic that picks "top of waitlist" misbehaves.

**2. `isDomainLead` is unscoped across the entire admin surface**

`lib/roles.ts:60-63` — `isDomainLead(userId)` returns true if the user has any `DomainLeadAssignment`, regardless of which domain the current request concerns. This is used as the auth gate in:

- `api.delibs.$id.ts` action (close, PATCH) — any domain lead can close/edit any domain's delibs.
- `api.domain-applications.$id.decisions.ts` — any domain lead can create Draft/Final decisions on applications for domains they don't lead.
- `api.domain-applications.$id.reviews.ts` — same pattern, plus no check that the assigned `cycleReviewerId` matches the DA's domain.
- `api.reviews.$id.submit.ts` / `.unsubmit.ts` — any domain lead can submit or unsubmit reviews in any cycle.
- `api.cycles.$cycleId.reviewers.ts` / `.$reviewerId.ts` — cross-domain reviewer assignment/removal.
- `api.cycles.$cycleId.interviewers.ts` — cross-domain interviewer assignment/removal.

This is the same pattern we'd want to enforce consistently: a domain lead should only mutate records whose domain matches one of their `domainLeadAssignments`. Either add a domain-scoped helper (`isDomainLeadFor(userId, domainId)`) and plumb it through every admin endpoint, or derive the target domain from the route params and check membership.

**3. `available-slots` "invited" check matches superseded decisions**

`api.cycles.$cycleId.available-slots.ts:17-27` uses `decisions: { some: { type: "InvitedToInterview" } }` to decide whether an applicant who isn't a cycle-member can read slots. Because `Decision` is append-only (per spec) and this PR's close-delibs change writes Final InvitedToInterview rows that then get superseded by later Accept/Reject decisions, `some:` still matches. An applicant who was briefly invited and then rejected retains slot-read access for any domain they query. Filter to the latest Released Decision per DA and only allow if its type is `InvitedToInterview`.

**4. Cycle can never transition to Completed if any interview was cancelled**

`api.cycles.$cycleId.status.ts:124-125` counts "pending" interviews as `status != "Completed"`. That includes `CancelledByApplicant` and `CancelledByAdmin` — which are audit rows, not in-progress interviews. Any realistic cycle ends with some cancelled-and-not-rebooked interviews; the count stays positive forever; the normal "Complete Cycle" action is blocked indefinitely, forcing hiring leads to use the `Force` escape hatch as the routine finish path. Gate on `status === "Scheduled"` only.

**5. Google OAuth doesn't check `email_verified`**

`lib/oauth.ts:314-327` verifies the ID token signature (#73's fix — good) but then reads `payload.email` without checking `payload.email_verified`. Google ID tokens include this boolean to indicate whether Google has confirmed the user owns the email in the token. Standard OAuth defense:

```js
if (!payload.email || !payload.email_verified) {
  throw new OAuthError("server_error", "Email not verified by Google");
}
```

Low practical impact today since `@dali.dartmouth.edu` / `@dartmouth.edu` Workspace emails are always verified, but it matters for the external-partner branch and any future non-Workspace flows.

**6. Delibs PATCH implements the exact anti-pattern the spec warns against**

`api.delibs.$id.ts:36-48`:

```ts
if (request.method === "PATCH") {
  const { columnOrder } = body;
  if (!columnOrder) return ...;
  await prisma.delibsSession.update({
    where: { id: params.id },
    data: { columnOrder },   // full-blob replacement
  });
  ...
}
```

Client sends the entire `columnOrder` JSON; server replaces the stored value. Two reviewers moving different cards simultaneously both read current state, both apply local changes, both PATCH — the second overwrites the first, silently losing one move.

The spec explicitly predicts this bug and prescribes a patch-based move API. From **`CYCLE_REDESIGN_SPEC.md:717-728`** ("Delibs Kanban Card Moves"):

> **Race condition:** Two reviewers simultaneously move different cards on the same board. Both read `DelibsSession.columnOrder`, apply their change locally, and write the full JSON back. The second write overwrites the first, silently dropping one person's move.
>
> **Plan:** Replace full `columnOrder` replacement with **patch-based move operations**. The client never sends the full `columnOrder` — it sends only what changed:
>
> - **API:** `POST /api/delibs-sessions/:id/moves { cardId: string, toColumn: string, position?: number }`
> - **Server:** inside a transaction, reads the latest `columnOrder`, removes the card from whichever column currently holds it, inserts it at the requested position (or appends to end of column), writes back.

Please implement the move API the spec describes.

**7. `/api/my-interview/cancel` and `/reschedule` pick an arbitrary interview** (related to round-3 item #9 on per-domain portal rendering)

`api.my-interview.cancel.ts:17-22` and `api.my-interview.reschedule.ts:31-45` both call `prisma.interview.findFirst({ where: { domainApplication: { application: { userId } }, status: "Scheduled" }})` with no `domainApplicationId` parameter and no `orderBy`. An applicant with Scheduled interviews in multiple domains gets an arbitrary one returned by Postgres. The portal's per-card "Cancel" button may cancel the *other* domain's interview. Reschedule is worse: it then passes the user's full `applicantDomainIds` list into `assignInterviewers`, so the rescheduled interview can be assigned under a different domain's interviewer pool than the one being rescheduled.

Accept `domainApplicationId` in the request body, scope the query to it, and verify the DA belongs to the requesting user. This is the same root theme as the portal's single-primary-stage issue: the applicant has multiple DomainApplications and the backend needs to stop pretending they have one.

Kiran-Jones (approved):
Looks good for now. Merging and making comments into gh issues

PR #99, opened 2026-04-19
Fix #97: extract shared selectHeroCycle helper

### LINKED ISSUE #97: Admin: extract hero-cycle selection helper to avoid duplicated logic
Tracked from PR #71 review (round 1, item #3).

`admin.tsx:137-143` and `admin.tsx:189-195`. `ActiveCycleHero` computes which cycle to feature at the top, and `PastCycles` re-computes the same expression to exclude it. If the hero-selection rule changes (e.g. prefer `UnderReview` over `Open`, or ignore stale Drafts) and only one site is updated, the active cycle will either be shown twice or hidden from both lists.

### Fix
Extract into a shared helper (e.g., `selectHeroCycle(cycles)`) and call from both components.

ashnaghanate:
## Summary

Fixes #97 — `ActiveCycleHero` and `PastCycles` in `admin.tsx` had duplicated hero-selection logic (Open/UnderReview preferred over Draft). If only one site got updated when the rule changed, the active cycle would either be shown twice or hidden from both lists.

Extracted a single `selectHeroCycle(cycles)` helper used by both components.

## Test plan
- [ ] Hiring lead cycle list shows the active cycle as hero and it doesn't appear again in Past Cycles
- [ ] With only Draft cycles, the Draft shows as hero
- [ ] With no cycles, "No active hiring cycle" shows

Closes #97

### DIFF
--- a/dali-api/app/routes/admin.tsx
+++ b/dali-api/app/routes/admin.tsx
@@ -132,15 +132,24 @@ export default function AdminDashboard() {
   );
 }
 
+// Single source of truth for hero cycle selection.
+// ActiveCycleHero features the result; PastCycles excludes it.
+// Prefers Open/UnderReview (truly active) over Draft.
+function selectHeroCycle(cycles: any[]): any | undefined {
+  return (
+    cycles.find((c: any) => {
+      const status = c.statusUpdates[0]?.newStatus;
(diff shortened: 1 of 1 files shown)


Kiran-Jones (approved):
LGTM

PR #110, opened 2026-04-19
Fix application submit by using useFetcher instead of raw fetch

### LINKED ISSUE #108: submit application button not working

Kiran-Jones:
Raw fetch couldn't detect React Router's server-side redirect, so the submit silently failed. Switching to useFetcher lets React Router handle the redirect automatically.

Closes #108


henryscheible:
Can you update the seed data so there is a cycle open by default?

henryscheible (approved):
LGTM

PR #114, opened 2026-04-20
Fix #113: block applicants from accessing member routes

### LINKED ISSUE #113: Applicants can see internal member UI by navigating to /reviewer
## Bug

Applicants (users with JWT `type: "applicant"`) can navigate to internal member routes like `/reviewer` and see member-facing UI (e.g. the "Review Dashboard" heading), even though the page also tells them they have no reviewer access. Applicants should not see any internal member UI at all.

## Root Cause

- `layout.tsx` (the parent layout for all member routes) calls `requireAuth` but never checks `auth.user.type`, so any authenticated user — including applicants — passes through.
- `home.tsx` blindly redirects all users to `/reviewer` without checking user type.

## Expected Behavior

Applicants navigating to any internal route (`/reviewer`, `/domain-lead`, `/admin-console`, etc.) should be redirected to `/portal`.

Kiran-Jones:
## Summary
- Adds a user type check in `layout.tsx` to redirect applicants to `/portal` before any member UI renders — protects all internal routes at once
- Makes `home.tsx` type-aware so applicants go to `/portal` instead of `/reviewer`

Fixes #113

## Test plan
- [ ] Log in as applicant → navigate to `/reviewer` → redirects to `/portal`
- [ ] Log in as applicant → navigate to `/admin-console` → redirects to `/portal`
- [ ] Log in as member → navigate to `/reviewer` → shows dashboard (unchanged)
- [ ] Unauthenticated → navigate to `/` → redirects to `/login` (unchanged)

### DIFF
--- a/dali-api/app/routes/home.tsx
+++ b/dali-api/app/routes/home.tsx
@@ -1,6 +1,10 @@
 import { redirect } from "react-router";
+import { requireAuth } from "~/lib/auth";
 import type { Route } from "./+types/home";
 
-export function loader({}: Route.LoaderArgs) {
+export async function loader({ request }: Route.LoaderArgs) {
+  const auth = await requireAuth(request);
+  if (!auth.ok) return redirect("/login");
+  if (auth.user.type === "applicant") return redirect("/portal");
   return redirect("/reviewer");
 }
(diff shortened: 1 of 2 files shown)


henryscheible (approved):
LGTM

PR #136, opened 2026-04-23
Auto-create ApplicationReview so feedback editors are editable

### LINKED ISSUE #135: Review comments cannot be edited
Image

There is no save button here and the text boxes are disabled so it is unclear how to proceed

ashnaghanate:
Closes #135

## Summary
When a reviewer opened a review page where no `ApplicationReview` row existed yet, the **Internal Feedback** and **Rejection Rationale** fields fell back to a disabled textarea with a "Save the review first to enable collaborative editing" placeholder — but the page has no save button for those fields (they persist via collab sync), leaving the reviewer stuck.

The loader now upserts an `ApplicationReview` when the current user has a `CycleReviewer` assignment for a matching domain on the application. Idempotent via the existing `(cycleReviewerId, domainApplicationId)` unique constraint.

## Test plan
- [ ] As a reviewer whose `ApplicationReview` was auto-assigned by the domain lead, open an assigned application — feedback/rejection rationale editors should render editable
- [ ] As a reviewer whose domain lead assigned them but somehow no review row exists, open the application — the row is created on load and both editors render editable
- [ ] Open the same application in two tabs simultaneously — no unique-constraint crash
- [ ] As a non-reviewer user with cycle access (e.g. hiring lead), the editors still render as before (disabled) since no matching `CycleReviewer` exists


Kiran-Jones (approved):
LGTM

PR #159, opened 2026-04-25
Remove vestigial showForRoles logic

Kiran-Jones:
## Summary
- Removes `showForRoles` state, `toggleRole` helper, and the "Conditional" badge from `ChallengeBuilder.tsx`
- Removes `showForRoles` field from the `Question` type in `types.ts`
- This was dead code: state and logic existed but no UI ever called `toggleRole`, so role-conditional visibility was never editable from the builder
- The feature is also redundant — domain-specific challenge forms already scope questions per domain, making `showForRoles` unnecessary

## Test plan
- [ ] Verify challenge builder still loads and functions (add/edit/delete questions)
- [ ] Verify no references to `showForRoles` or `toggleRole` remain in the codebase

### DIFF
--- a/dali-api/app/components/ChallengeBuilder.tsx
+++ b/dali-api/app/components/ChallengeBuilder.tsx
@@ -11,7 +11,6 @@ export interface BuildQuestionInput {
   description?: string
   optionsText?: string
   accept?: string
-  showForRoles?: string[]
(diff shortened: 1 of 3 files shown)


henryscheible (approved):
LGTM

PR #190, opened 2026-04-26
Refactor email templates to rubric-pattern + per-cycle decision bindings

henryscheible:
## Summary

- New `EmailTemplate` / `EmailTemplateVersion` / `CycleDecisionEmail` trio mirroring `Rubric` / `RubricVersion` / `DomainApplicationCycle`. Templates are named, versioned objects independent of cycles; per-cycle bindings (one per `DecisionType`) determine which template version fires when a decision is released.
- Decision-release flow now looks up the binding by `(applicationCycleId, decisionType)` instead of doing a global `findFirst` by enum type. When no binding exists, the release proceeds but the email send is skipped with a clear `console.warn` (same swallow-and-continue posture, no longer silent).
- New hiring-lead UI: `/email-templates` library (list + create + append versions, mirrors `/rubrics`), and a "Decision Emails" picker section on the cycle admin page (one row per `DecisionType`, mirrors `GeneralRubricPicker`). Picker locks once a decision of that type has been Released; empty state explicitly warns "no template assigned — releasing this decision will not send an email."
- `/emails` (batch send) is rewritten as send-only: pick a template from the new library, pick a recipient group, send. Editor moved out to `/email-templates/$id`.
- `ApplicationReceived` (auto on submit) and `InterviewInviteMentor` (manual batch only) keep using the legacy global-by-type lookup, now via the renamed `LegacyEmailTemplate` model. `RejectedPostInterview` and the pre/post-interview discriminator are intentionally left for a follow-up.

## What this addresses

PR #112 made email *content* DB-backed and versioned but left the type taxonomy hardcoded in two TS arrays and made templates fully global — every cycle used the same templates with no way to vary copy without overwriting history. This PR puts emails on the same footing as rubrics/challenges. The `EmailTemplate` model has no `type` field at all; the slot a template fills is determined externally by the binding row, exactly as `Rubric` works.

## Migration ⚠️

`prisma/migrations/20260426170000_email_template_rubric_pattern/migration.sql` is **data-moving**:
- `EmailTemplate` table renamed to `EmailTemplateLegacy` (constraints + index renamed too).
- New tables created.
- Backfill: each unique legacy `type` becomes one `EmailTemplate` parent (deterministic id `tmpl_<lowercased_type>`), legacy rows become `EmailTemplateVersion` rows preserving id/createdAt/version/subject/body/createdById, and every cycle whose latest status is `Open` or `UnderReview` gets `CycleDecisionEmail` rows for the 4 `DecisionType`s pointing at the highest-versionNumber matching template. `ON CONFLICT DO NOTHING` so the backfill is idempotent.
- No data loss — `LegacyEmailTemplate` is preserved (still used by `ApplicationReceived` and `InterviewInviteMentor`). Drop in a follow-up after one prod deploy verifies the new flow.

`migration-check.yml` will run pgfence — please review the analysis.

## Out of scope (deferred)

- Drop `LegacyEmailTemplate` and shrink `EmailTemplateType` enum.
- `RejectedPostInterview` wiring + pre/post-interview discriminator (needs a `Decision` schema change or runtime interview-history check).
- Per-cycle binding for `ApplicationReceived` / `InterviewInviteMentor`.
- Surfacing email-send failures in the UI (still silently swallowed today).

## Unrelated finding

`dali-api/prisma.config.ts:7` has `directUrl: process.env.DIRECT_URL`, but Prisma 7's `Datasource` type only accepts `url` and `shadowDatabaseUrl` — TS rejects the field, and Prisma silently ignores it at runtime. So the `DIRECT_URL` Fly secret currently set on `dali-api-dev` and `dali-api-prod` (intended to dodge the P1002 advisory-lock issue against the Neon pooler) is **not actually consumed by Prisma**. Needs a separate fix — likely move `directUrl` into the `datasource db` block in `schema.prisma`.

## Test plan

- [ ] CI: `test.yml` Vitest passes (172 unit tests local incl. 13 new — covering binding-present send, no-binding skip+warn, 409 on non-Final, 403 on non-hiring-lead, netId@dartmouth.edu fallback, sendEmail failure swallowed)
- [ ] CI: `test.yml` Playwright E2E passes against the seeded Postgres
- [ ] CI: `migration-check.yml` clean (pgfence)
- [ ] CI: `build-check.yml` clean
- [ ] Preview deploy: visit `/email-templates`, create "Rejected — Standard" + "Rejected — Warm" with one version each
- [ ] Preview deploy: on the active cycle's admin page, bind `Rejected` slot to "Standard"; release a Rejected decision and confirm the standard copy lands in the dev Gmail inbox
- [ ] Preview deploy: rebind `Rejected` to "Warm", release another Rejected decision, confirm the warm copy fires
- [ ] Preview deploy: leave `Accepted` unbound, release an Accepted decision, confirm the release succeeds and a `No email template bound` warning appears in logs (no send)
- [ ] Preview deploy: lock guard — after a Released decision of a given type, the picker for that slot shows "(locked — decisions of this type already released)"


Kiran-Jones (approved):
LGTM - tested on the deployed link

PR #212, opened 2026-04-27
Fix #204: add admin UI to create and delete domains

### LINKED ISSUE #204: Add UI that allows admins to create and remove domains
Currently there is no UI that allows domains to be created. Make one, and restrict deletions if the domain is in use in any other way. Have it be accessible to admins only (NOT hiring leads).

henryscheible:
Closes #204.

## Summary
- Add a **Domains** section to `/admin-console` (admin-only) with a create form and per-row Delete. Delete is disabled when the domain is referenced by any of its seven Prisma relations and the row shows what is keeping it alive (e.g. *In use by 2 application cycles, 3 rubrics*).
- Tighten `POST /api/domains` from `isHiringLead` to `isAdmin`. The loader stays open to hiring leads / domain leads / admins (other UIs depend on it) and now returns `_count` for each domain so callers can decide whether deletion is allowed.
- Add `DELETE /api/domains/:domainId` (admin-only). Inside a transaction it counts references across all seven `Domain` relations; non-zero → `409` with a message naming the blocking relation(s); otherwise the row is deleted.
- Mirror the create / delete logic in the page action (`create-domain`, `delete-domain` intents) so the Domains section uses the existing fetcher pattern.

## Implementation notes
- "In use" = referenced by any of: `challengeVersions`, `applicationCycles`, `domainLeadAssignments`, `cycleReviewers`, `cycleInterviewers`, `rubrics`, `delibsSessions`. Any reference is blocking — including `domainLeadAssignments`. Per the plan, if you'd like lead assignments to cascade-clean instead of block, that's a follow-up.
- No Prisma schema migration. No CRDT-touching changes.
- The existing `/admin-console` route guard still admits admins or hiring leads, so the Members table stays visible to HLs. The Domains section is hidden when `!isAdmin`.

## Tests
- New Vitest unit test (`api.domains.delete.test.ts`) covers: non-admin → 403, non-DELETE → 405, missing domain → 404, in-use → 409 with relation names, no references → delete.
- New Playwright spec (`e2e/domain-management.spec.ts`) covers: admin creates a fresh domain and deletes it; Delete is disabled and labeled "In use by …" for the seeded `Engineering` domain; hiring leads do not see the Domains section.

## Test plan
- [ ] CI: `test.yml` (Vitest + Playwright) green
- [ ] CI: `build-check.yml`, `migration-check.yml`, `codeql.yml` green
- [ ] Manual: as admin, create + delete a domain; confirm Delete is disabled for any seeded in-use domain with a tooltip listing the blocking relations
- [ ] Manual: as a hiring lead (non-admin), confirm the Domains section is not rendered

### DIFF
--- a/dali-api/app/routes.ts
+++ b/dali-api/app/routes.ts
@@ -50,6 +50,7 @@ export default [
 
   // Domain & member management API
   route("api/domains", "routes/api.domains.ts"),
+  route("api/domains/:domainId", "routes/api.domains.$domainId.ts"),
   route("api/domains/:domainId/leads", "routes/api.domains.$domainId.leads.ts"),
   route("api/members", "routes/api.members.ts"),
   route("api/members/:memberId/roles", "routes/api.members.$memberId.roles.ts"),
--- a/dali-api/app/routes/__tests__/api.domains.delete.test.ts
+++ b/dali-api/app/routes/__tests__/api.domains.delete.test.ts
@@ -1,0 +1,110 @@
+import { describe, it, expect, beforeEach, vi } from "vitest";
+
+vi.mock("~/lib/db");
+vi.mock("~/lib/auth");
+vi.mock("~/lib/roles");
+
+import { prisma } from "~/lib/db";
+import { requireAuth } from "~/lib/auth";
+import { isAdmin } from "~/lib/roles";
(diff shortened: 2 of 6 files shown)


Kiran-Jones (approved):
LGTM

PR #465, opened 2026-05-09
Add configurable deadline extension with applicant-facing notice

Kiran-Jones:
## Summary
- Hiring leads can extend a cycle's close date by a configurable amount (default 48 hours, unit dropdown for hours/days) from the cycle setup tab. Click-to-confirm so a stray click doesn't push the deadline.
- New nullable `originalCloseDate` column on `ApplicationCycle` captures the pre-extension deadline on the first extension and is preserved across subsequent extensions.
- Applicant portal renders a `Deadline extended` pill on the deadline line during the window between the original and current close dates, so a returning applicant immediately sees why the date is later than they remember.

## Implementation notes
- Manually setting the close date via the existing date picker clears `originalCloseDate` — treated as a fresh deadline, not an extension.
- Migration is purely additive: `ALTER TABLE "ApplicationCycle" ADD COLUMN "originalCloseDate" TIMESTAMP(3)`. Nullable, no backfill.
- The action validates that the amount is a positive number, that the unit is `hours` or `days`, and that a `closeDate` is already set (refuses to "extend" nothing). Auth and hiring-lead role checks reuse the existing guard at the top of the action.
- No changes to the auto-close logic in `app/hiring/lib/cycles.ts` — it operates on `closeDate` and continues to work unchanged.
- The extension pill is computed client-side inside `DeadlineLine`'s `useEffect` to avoid SSR/CSR mismatches (the existing `formatDeadline` already uses this pattern).

## Test plan
- [x] `npx vitest run app/hiring/routes/__tests__/lead.cycle.$id.test.ts` — 6 new tests covering: first extension captures `originalCloseDate`, subsequent extensions preserve it, manual `set-close-date` clears it (set + unset), non-positive amount rejected, no-close-date rejected
- [x] Wider related suites (cycles, status, modal, lead.cycle action) — 42 tests pass total
- [ ] Manual smoke: as hiring lead, set a close date, click Extend with default 48 / hours, see Original deadline text appear; verify portal shows the pill while now is between original and new close
- [ ] Manual smoke: extend twice, confirm `originalCloseDate` stays anchored to the very first deadline (pill window still spans original → final close)
- [ ] Manual smoke: edit close date via the picker, confirm pill disappears (originalCloseDate cleared)

## Migration safety
Additive nullable column — `migration-check.yml` / `pgfence` should pass. No data-loss risk.

### DIFF
--- a/dali-api/app/hiring/routes/__tests__/lead.cycle.$id.test.ts
+++ b/dali-api/app/hiring/routes/__tests__/lead.cycle.$id.test.ts
@@ -36,6 +36,10 @@ beforeEach(() => {
   mockPrisma.rubricVersion = { findUnique: vi.fn() };
   mockPrisma.applicationReview = { count: vi.fn() };
   mockPrisma.domainApplicationCycle = { upsert: vi.fn().mockResolvedValue({}) };
+  mockPrisma.applicationCycle = {
+    update: vi.fn().mockResolvedValue({}),
+    findUnique: vi.fn(),
+  };
 
   vi.mocked(requireAuth).mockResolvedValue({
     ok: true,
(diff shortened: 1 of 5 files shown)


Kiran-Jones:
Pulled in `dev` and resolved the UI conflict in `lead.cycle.$id.tsx` — kept the timezone subhead and ET-zoned date input from #464 and layered the extension form + original-deadline notice on top.

Fixed the auto-close issue you saw on the preview: extending a cycle whose `closeDate` had already passed left it materialized as `UnderReview`, so the portal kept showing the closed view even though the new deadline was in the future. The `extend-close-date` action now writes a new `Open` status update in the same transaction when (a) the latest status is `UnderReview` and (b) the new `closeDate` is in the future. Skipped for `Completed` (terminal) and `Draft` (cycle never opened — bumping the date is enough).

New regression test added (`reopens the cycle when extending past now after auto-close`) plus a paired `does not reopen when the new close date is still in the past` to prove the gate. 44 related tests pass.

Kiran-Jones:
Pushed three fixes addressing the testing feedback:

**Bug fix — set-close-date also reopens.** This was the actual cause of what you saw on the preview. The reopen logic was only on the Extend button. If you used the date picker to set a future date on an auto-closed cycle, it stayed in `UnderReview` and the portal kept showing the closed view. Both actions now share a `reopenIfNeeded` helper that runs inside the same transaction. Two new tests cover this: `reopens the cycle when manually setting a future date on a closed cycle` and `does not reopen on set-close-date when status is already Open`.

**Confirmation modal for Extend.** Replaced the inline two-click confirm with a real modal that shows:
- Current close (in ET)
- New close after extension (in ET)
- A warning if the cycle will reopen as a result of the extension
- A red error if the new close time would still be in the past

**Post-action notice banner.** The action now redirects with `?notice=...` and a new `CloseDateNotice` component reads it on mount, shows a dismissible banner, and strips the query string. Distinct tones for "saved" vs "saved and reopened" so you immediately see whether the applicant view changed — no more bouncing to the portal to check.

47 related tests pass. If you re-test on the preview, the flow that was broken (`set close date to yesterday` → `extend 48h`) now reopens correctly because of fix #1, and you'll see a `Deadline extended and applications reopened` banner confirming it.

Kiran-Jones:
Restructured the close-date card to match how a hiring lead actually thinks about this:

**Section 1 — Application Close Date.** The picker shows the *original/intended* close date (5/8 in your example). If an extension is active and you save a new picker date, the extension delta moves with it — original and effective close shift together by the same amount.

**Section 2 — Extension.** A single state-style form with the current amount preselected. Default is 48 hours when no extension exists. Save replaces (doesn't stack) — saving 48h twice in a row is idempotent. A "Remove extension" link snaps the close date back to the original.

**Section 3 — Effective Close.** Read-only display of when applications actually stop, in ET, with a sentence describing the extension window.

The extension confirm modal now shows "Original close → New effective close" instead of conflating the two, so the preview matches the section labels.

So your test scenario now reads cleanly:
- Save 5/8 → picker stays at 5/8, extension is empty, effective close = 5/8
- Set extension to 48h → picker still 5/8, extension shows "48 hours active", effective close = 5/10
- Page reload → same view, no confusion

50 tests pass. Pushed to the branch — preview should pick it up shortly.

PR #474, opened 2026-05-09
Hide deselected domain applications and validate required questions on submit

Kiran-Jones:
## Summary

Investigating a prod report of an application showing an "empty" domain
application surfaced two related correctness gaps in the apply →
hiring-review path. This PR closes both.

### 1. Hide deselected domains from hiring views

When an applicant deselects a domain mid-application, the server
intentionally keeps the `DomainApplication` row but flips
`selected = false` so answers survive a re-select (per the schema
comment on `DomainApplication.selected`). The applicant-facing loaders
(`portal`, `portal.application`) and `analytics` already filter on
`selected: true`, but four hiring-side loaders did not, so deselected
domains kept appearing in the lead/reviewer UIs:

- `app/hiring/routes/lead.cycle.$id.tsx`
- `app/hiring/routes/domain-lead.tsx`
- `app/hiring/routes/domain-lead.delibs.$id.tsx`
- `app/hiring/routes/reviewer.application.$id.tsx`

All four now scope `domainApplications` to `selected: true`.

### 2. Server-side required-question validation on submit

The submit action only validated word limits server-side; required
questions were enforced solely by the client's review-modal gate. That
means a selected-but-empty domain could reach the DB through a stale
client state, a manual replay, or a future client refactor that loosens
the gate — which is consistent with the prod row that prompted this
investigation (the affected domain has required questions, so the gate
*should* have caught it).

The submit action now re-checks every selected domain (and the general
form) against the required questions on the picked `ChallengeVersion`,
returning a generic `error` field that the client surfaces via alert.
This is defense-in-depth: the client gate continues to be the primary
UX, and submissions that satisfy it stay unaffected.

## Test plan

- [x] `npm test` — full suite (645 → 647 tests, all passing)
- [x] Two new unit tests in `portal.apply.test.ts` cover:
  - selected domain with empty `domainAnswers` payload → rejected
  - selected domain omitted from `domainAnswers` entirely → rejected
- [ ] Manual: deselect a partially-filled domain, submit with another, confirm the deselected one no longer appears in the lead cycle view, domain-lead view, delibs board, or reviewer view.
- [ ] Manual: leave a required field blank in a selected domain, attempt to bypass the client gate (e.g., DevTools), confirm the server rejects with a clear message.

## Notes for the existing prod row

This fix is forward-only. Existing rows in the offending state will be
hidden from the hiring UI by change (1) if the applicant in question
deselected the domain — which is the most likely path. If the row has
`selected: true` with empty answers, it'll need a one-off DB cleanup;
happy to follow up with a query once we look at the actual row.

### DIFF
--- a/dali-api/app/hiring/routes/domain-lead.delibs.$id.tsx
+++ b/dali-api/app/hiring/routes/domain-lead.delibs.$id.tsx
@@ -52,6 +52,7 @@ export async function loader({ request, params }: Route.LoaderArgs) {
 
   const domainApplications = await prisma.domainApplication.findMany({
     where: {
+      selected: true,
       challengeVersion: { domainId: session.domainId },
       application: {
         applicationCycleId: session.applicationCycleId,
--- a/dali-api/app/hiring/routes/domain-lead.tsx
+++ b/dali-api/app/hiring/routes/domain-lead.tsx
@@ -85,7 +85,7 @@ export async function loader({ request }: Route.LoaderArgs) {
               user: true,
               statusUpdates: { orderBy: { createdAt: "desc" }, take: 1 },
               domainApplications: {
-                where: { challengeVersion: { domainId: assignment.domainId } },
+                where: { selected: true, challengeVersion: { domainId: assignment.domainId } },
                 include: {
                   challengeVersion: { include: { domain: true } },
                   reviews: {
--- a/dali-api/app/hiring/routes/lead.cycle.$id.tsx
+++ b/dali-api/app/hiring/routes/lead.cycle.$id.tsx
@@ -130,6 +130,7 @@ export async function loader({ request, params }: Route.LoaderArgs) {
           user: true,
           statusUpdates: { orderBy: { createdAt: "desc" }, take: 1 },
           domainApplications: {
+            where: { selected: true },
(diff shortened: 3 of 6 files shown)


Kiran-Jones:
Follow-up commit (7d6178b) — closes the by-id bypass on the review-pipeline write paths so deselected DAs can't be assigned reviewers, scored, scheduled for interviews, or released-to even if a stale id reaches the API.

Hardened endpoints:
- `api.cycles.$cycleId.domains.$domainId.auto-assign.ts` — bulk reviewer assignment now scopes `findMany` to `selected: true`.
- `api.domain-applications.$id.reviews.ts` — manual reviewer assignment 404s on a deselected DA.
- `api.domain-applications.$id.decisions.ts` — decision create 404s on a deselected DA.
- `api.domain-applications.$id.schedule-interview.ts` — applicant interview booking 404s on a deselected DA.
- `api.decisions.$id.release.ts` — decision release 404s on a deselected DA.

Test mocks updated to set `selected: true` explicitly so the assumption is no longer implicit. Full suite: 647/647 passing.

Caveat (deliberately out of scope): this is forward-only. Any reviews / interviews / decisions that were already created against a now-deselected DA before this PR will still exist in the DB. Those need a one-off cleanup once we identify the affected rows.

PR #492, opened 2026-05-12
Drop termId from DomainLeadAssignment

sophp516:
## Summary
- Drops `DomainLeadAssignment.termId` column, its `Term` FK, and the `(memberId, domainId, termId)` composite unique. Replaces with `(memberId, domainId)` unique.
- Removes `getCurrentTermId()` calls from three admin/console route actions.
- Seed updated.

## Why
The expansion_v0 migration added `termId NOT NULL` to an existing populated table, which crashes on staging deploy:
> ERROR: column "termId" of relation "DomainLeadAssignment" contains null values

The deeper issue is that no read path filters DomainLeadAssignment by termId. The only place `termId` was set was at create time, via a synthesized "current term" lookup. Dropping the column simplifies the model: an assignment persists until manually removed.

## Migration safety
The migration drops a NOT NULL FK column. On dev/staging/prod:
- **Dev** local: applies cleanly (table empty).
- **Staging**: previous deploy failed and left no rows mid-migration. After this PR, redeploy should resolve the failed migration. If staging's `DomainLeadAssignment` has multiple rows for the same `(memberId, domainId)` across distinct terms, the new unique index will fail; dedupe first.
- **Prod**: same caveat — check for `(memberId, domainId)` duplicates before promoting.

## Test plan
- [ ] `cd dali-api && npm run typecheck`
- [ ] `cd dali-api && npm test` (unit)
- [ ] `cd dali-api && npm run test:e2e` (covers admin-console + domain-lead role checks)
- [ ] On staging: after deploy, log in as admin and verify domain leads are listed; add/remove a lead through `/admin-console/members` and `/admin-console/domains`.

### DIFF
--- a/dali-api/app/routes/admin-console.domains.tsx
+++ b/dali-api/app/routes/admin-console.domains.tsx
@@ -3,7 +3,6 @@ import { redirect, useLoaderData, useFetcher } from "react-router";
 import type { Route } from "./+types/admin-console.domains";
 import { prisma } from "~/lib/db";
 import { requireAuth, withAuth } from "~/lib/auth";
-import { getCurrentTermId } from "~/lib/terms";
 import { isAdmin } from "~/lib/roles";
 import { describeDomainUsage } from "./api.domains.$domainId";
 import { ChevronDown, Trash2, Plus } from "lucide-react";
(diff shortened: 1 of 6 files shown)


Kiran-Jones (approved):
LGTM

PR #495, opened 2026-05-12
Strip DomainLeadAssignment.termId from expansion_v0; idempotent drop

Kiran-Jones:
## Summary
- Staging's Fly Deploy after #492 still failed at `20260511000000_expansion_v0_schema` with the same SQLState 23502: the migration tries `ALTER TABLE \"DomainLeadAssignment\" ADD COLUMN \"termId\" TEXT NOT NULL`, which Postgres rejects when the table already has rows. Staging Neon is rebuilt from prod each deploy, so the broken migration re-runs and aborts before #492's follow-up drop can execute.
- Remove the four lines in `expansion_v0` that introduce `termId` on `DomainLeadAssignment` (ADD COLUMN, `(domainId, termId)` index, composite unique, FK to Term). The column is never created.
- Rewrite `20260512023900_drop_domain_lead_assignment_term` with `IF EXISTS` / `IF NOT EXISTS` so its drops no-op on the now-clean chain while still creating the `(memberId, domainId)` unique + `(domainId)` index that `schema.prisma` already declares.

## Why hand-edit applied migrations
CLAUDE.md normally forbids this. Justified per project notes:
- staging/dev Neon are rebuilt every deploy → checksum drift in those envs is harmless
- prod has never run either migration → no production state to corrupt
- the alternative (Term backfill + seed) expands scope the merged #492 deliberately rejected (DomainLeadAssignment is no longer term-scoped)

`migration-check.yml` will flag this PR. That's expected; the failure should be ack'd, not worked around with `--no-verify`.

## Test plan
- [ ] CI: `test.yml` (unit + Playwright) passes
- [ ] CI: `build-check.yml` passes
- [ ] CI: `migration-check.yml` — expected to fail on the hand-edit; review and ack
- [ ] After merge → promote to staging: Fly Deploy `release_command` (prisma migrate deploy) completes without 23502 or P3009
- [ ] Verify on staging DB: `DomainLeadAssignment` has no `termId` column, has `(memberId, domainId)` unique and `(domainId)` index

### DIFF
--- a/dali-api/prisma/migrations/20260511000000_expansion_v0_schema/migration.sql
+++ b/dali-api/prisma/migrations/20260511000000_expansion_v0_schema/migration.sql
@@ -92,9 +92,6 @@ ADD COLUMN     "displayName" TEXT,
 ADD COLUMN     "isInternProgram" BOOLEAN NOT NULL DEFAULT false;
 
 -- AlterTable
-ALTER TABLE "DomainLeadAssignment" ADD COLUMN     "termId" TEXT NOT NULL;
-
--- AlterTable
 ALTER TABLE "User" ADD COLUMN     "bioDocId" TEXT,
 ADD COLUMN     "classYear" INTEGER,
 ADD COLUMN     "githubUrl" TEXT,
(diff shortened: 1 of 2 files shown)


Kiran-Jones:
Superseded — switching to a full revert of the expansion-v0 schema (kept calendar additions only).

PR #507, opened 2026-05-12
Fix annotation popup getting clipped off-screen near viewport edges

sophp516:
## Summary
- The annotation popup in the application review uses `position: fixed` but added `window.scrollY` to its top coordinate, so once the reviewer scrolled the popup was placed far below the viewport — invisible.
- Horizontal clamping only guarded the right edge (`Math.min(...)`), so the left side and small viewports could still clip it.
- Now: coordinates are computed in viewport space (no `scrollY`), both axes are clamped to the viewport with an 8px margin, and the popup flips above the selection when there isn't room below it.

## Test plan
- [ ] Select text near the **bottom** of the viewport after scrolling — popup appears above the selection, fully visible.
- [ ] Select text near the **top** of the viewport — popup appears below as before.
- [ ] Select text near the **left edge** — popup doesn't get clipped on the left.
- [ ] Select text near the **right edge** — popup doesn't get clipped on the right (regression check; this case was already partially handled).
- [ ] Click an existing annotation in each of the above positions — edit popup obeys the same clamping.
- [ ] Resize the window narrow (~600px) — popup stays inside the viewport.

### DIFF
--- a/dali-api/app/hiring/components/ApplicationViewer.tsx
+++ b/dali-api/app/hiring/components/ApplicationViewer.tsx
@@ -30,10 +30,29 @@ interface Annotation {
 interface Popover {
   x: number
   y: number
+  anchorTop: number
   start: number
   end: number
   fieldKey: string
   annotationId?: string
+}
+
+const POPOVER_WIDTH = 256
+const POPOVER_ESTIMATED_HEIGHT = 220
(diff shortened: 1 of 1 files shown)


Kiran-Jones (approved):
LGTM

PR #515, opened 2026-05-14
Replace JWT auth with opaque session cookies (closes #407)

### LINKED ISSUE #407: Replace JWT access/refresh token auth with opaque session cookies
## Proposal

Replace the current JWT access token + rotating refresh token scheme (`app/lib/auth.ts`, `app/lib/oauth.ts`) with a single opaque session id stored in an HttpOnly cookie, backed by a `sessions` table in Postgres.

## Background

Today's auth is the textbook OAuth-style two-token design:

- **Access token (AT)**: 15-min JWT, verified statelessly via HMAC on every request (`verifyAccessToken`).
- **Refresh token (RT)**: long-lived opaque string in DB, single-use / rotated on each refresh, with family-level reuse detection.
- **Silent refresh**: `requireAuth` transparently trades an expired AT for a new AT/RT pair when an RT cookie is present, and `withAuth` glues the resulting `Set-Cookie` headers onto the outgoing response.
- **Hard cap**: 30-day absolute session lifetime added in `2c01467` to bound RT chain length.

This works, but it's a lot of moving parts for what we actually need.

## Why we're considering the change

The strongest argument for the AT/RT split is **stateless validation on the hot path** — verify a JWT signature, no DB hit, scale forever. That argument doesn't really land here:

- Every protected route already calls `isAdmin`, `isHiringLead`, `hasCycleAccess`, etc. (`app/lib/roles.ts`), which hit the DB to re-fetch the user's roles. Authorization is already DB-backed; we're not actually skipping a roundtrip by making identity stateless.
- We're a same-origin web app with both tokens as HttpOnly cookies. The classic "AT leaks more than RT in side channels" argument (mobile clients, bearer headers, third-party resource servers) doesn't apply to our deployment.
- Hundreds of users, one Postgres. We are nowhere near the scale where one extra indexed lookup per request matters.

Meanwhile we're paying real complexity costs:

- `inFlightRefreshes` dedup map to prevent parallel loaders from tripping reuse detection on the same RT.
- `withAuth` must be threaded through every loader/action return path; forgetting it silently breaks the silent-refresh UX.
- Recent bug history: `1757c2a` (users bounced every 15 min because silent refresh wasn't wired up), `2c01467` (RT chains could live forever without an absolute cap).
- Two cookies, two expiries, two clear paths, family revocation, JWT_SECRET rotation story, etc.

## Proposed design

- **`sessions` table** keyed by random opaque id (32+ bytes from `crypto.randomBytes`, base64url-encoded).
  - Columns: `id` (PK, indexed), `userId` (FK), `createdAt`, `lastUsedAt`, `expiresAt`, `revokedAt`, plus optional `userAgent` / `ip` for audit.
  - Rolling 30-day `expiresAt` updated on use; hard cap at 30 days from `createdAt` to match current behavior.
- **One HttpOnly cookie** (`__dali_sid`), `SameSite=Lax`, `Secure` in prod.
- **`requireAuth(request)`** becomes: parse session id → single indexed `findUnique` → check `revokedAt` and `expiresAt` → return `{ user }`. No JWT verify, no AT/RT branches, no silent refresh, no `withAuth` glue. Loaders/actions just return their data.
- **Logout / revocation**: set `revokedAt`. Instant, no 15-min stale window.
- **\"Log out everywhere\"**: trivially `UPDATE sessions SET revokedAt = NOW() WHERE userId = ?`.
- **Role demotion**: already takes effect immediately today (authz is DB-backed), and continues to.

## Pros

- **Less code**: drop JWT signing/verifying, RT rotation, family revocation, `inFlightRefreshes`, `withAuth`, and the dual-cookie machinery. Probably nets ~hundreds of lines deleted.
- **One cookie, one code path** — no silent refresh dance, no chance of forgetting `withAuth` on a route.
- **Instant revocation** for logout, security incidents, and \"sign out all sessions.\"
- **No `JWT_SECRET` rotation problem.**
- **Easier to reason about and easier to onboard new contributors to.** Standard session pattern, no OAuth-specific concepts.
- **Better audit trail**: `sessions` rows are inspectable (last-used, IP, UA) in a way that stateless JWTs aren't.

## Cons / tradeoffs

- **One extra DB lookup per request** (one indexed PK lookup). In practice negligible at our scale and against an already-warm connection, but it is a real change.
- **Loses RT reuse detection.** With opaque session ids there's nothing analogous to \"this token was already rotated, alarm.\" Mitigation: short rolling expiry + audit logging on session use is enough for an internal tool with this threat model. We could add detection later if needed.
- **Migration work**: schema, code, and a cutover plan that doesn't log everyone out simultaneously (probably: dual-read for a release, then drop the JWT path).
- **Bearer-token API access** (`auth.ts:133-136`) goes away in its current form. Not used today, but if we ever want programmatic API access we'd add scoped API keys as a separate concern (which is the right design anyway — long-lived bearer JWTs are not what you want for API tokens).
- **Rewriting working auth is inherently risky.** Bugs in this layer are user-visible and security-sensitive.

## Why not just keep what we have

We talked through this and agreed: the current design is fine and we shouldn't churn for the sake of it. But:

- We just shipped two follow-up fixes to the JWT scheme in the last week (`1757c2a`, `2c01467`). The complexity is producing bugs.
- The justifications for the split (stateless hot path, side-channel asymmetry, scale) don't actually apply to this codebase.
- For a Postgres-backed internal tool with hundreds of users, the simpler design is also the more secure one (instant revocation > 15-min AT staleness).

We'd pick sessions if we were greenfielding this app today. The question is whether the migration cost is worth eating now vs. later, vs. never.

## Out of scope

- Changing the OAuth/CAS login flows themselves — those still issue a session at the end, just an opaque one instead of a JWT pair.
- API keys for programmatic access — separate issue if/when we need it.
- Touching `roles.ts` or any authorization logic.

## Open questions

- Cutover strategy: dual-read for one release, or hard cutover with forced re-login?
- Do we want `userAgent` / `ip` on the sessions row from day one, or add later?
- Should we ship a `sessions` admin view for users to see/revoke their own active sessions? (Nice-to-have, not blocking.)

Kiran-Jones:
## Summary

Replaces the current JWT access + rotating refresh token scheme with an opaque session id stored in a `Session` table, used uniformly for browser cookies (`__dali_sid`) and `Authorization: Bearer` headers (MCP-ready).

Background and design tradeoffs in [`SESSION_AUTH_PLAN.md`](./SESSION_AUTH_PLAN.md) (added in this PR). Closes #407.

## Why

- JWT statelessness wasn't earning its keep — every protected route already hits the DB for role checks, so the "save a roundtrip on the hot path" argument doesn't apply here.
- Recent bug history (`1757c2a`, `2c01467`) traced back to the silent-refresh complexity (`withAuth` glue + `inFlightRefreshes` dedup).
- The MCP track (`dali-os-mcp.md`) needed Bearer-token API access against the same auth surface. Unifying both onto one `Session` model is simpler than running two parallel systems.

## What changed

| Area | Effect |
|---|---|
| `lib/auth.ts` | 217 → 132 lines. `signAccessToken` / `verifyAccessToken` / `JWT_SECRET` / `inFlightRefreshes` / silent refresh / `withAuth` all gone. `requireAuth` is one indexed `findUnique` and preserves `auth.user.sub` shape. |
| `lib/cookies.ts` | Single `__dali_sid` HttpOnly cookie. Added `parseBearerHeader` + `parseSessionId` (cookie wins over header). |
| `lib/oauth.ts` | 337 → 230 lines. Removed `issueTokens`, `refreshTokens`, `revokeToken`, `createTokenPair`. PKCE / authorize-session surface preserved as foundation for the MCP track. |
| `lib/session.ts` (new) | `issueSession`, `lookupSession`, `rollSession`, `revokeSession`, `revokeAllForUser`, `revokeAllForGrant`. Raw ids never stored — `Session.id` = `sha256(raw)`. |
| Prisma schema | Drops `RefreshToken`, adds `Session` with nullable `grantId String?` (no FK yet — wired when `OAuthGrant` lands with the MCP track). |
| Routes | All three login flows + dev-login + logout + `/oauth/token` + `/oauth/revoke` rewired to the session model. `refresh_token` grant removed from `/oauth/token`. |
| Codemod | 529 `withAuth(auth, X)` calls stripped across 80 files; loaders/actions just return their data. |
| Collab | `collab/auth.ts` rewritten as session-backed verifier (returns legacy `{ sub }` shape); two loaders that pass `collabToken` to the Tiptap editor now use `parseSessionCookie`. |
| Tests | `auth.test.ts`, `cookies.test.ts`, `oauth.test.ts` rewritten. New `session.test.ts` with 15 cases including hash-at-rest assertions. 692 tests passing. |
| Cleanup | `jose` removed from `package.json`. `JWT_SECRET` removed from `docker-compose.yml` and `.env.example`. README + `dali-os-mcp.md` updated. |

## Cutover

**Hard cutover.** The migration drops `RefreshToken` (data-losing). Every active user gets logged out once and re-signs in on next request. Plan recommends scheduling outside hiring-cycle deadline windows and posting a heads-up in `#dali-eng` ~24h ahead — see [`SESSION_AUTH_PLAN.md` § Cutover plan](./SESSION_AUTH_PLAN.md).

## Risks specifically addressed during review

- **Collab/auth.ts** was not in the original plan; caught during implementation. Phase 5.5 added; without it Tiptap docs would stop authenticating mid-deploy.
- **`grantId` ships without an FK** — plain `String?` column until `OAuthGrant` lands in the MCP track. `revokeAllForGrant` is in place but no-ops until then.
- **`Session.id` is the SHA-256 hash** of the raw id (not a separate column) — PK lookup stays indexed, raw never hits the DB, mirrors the `RefreshToken.tokenHash` precedent.
- **`auth.user.sub` preserved** — the 111-file codemod is purely about removing `withAuth` wrappers, not renaming. Regression guarded by a Vitest assertion.

## Commits

| # | Commit | Theme |
|---|---|---|
| 1 | `docs:` | `SESSION_AUTH_PLAN.md`, MCP plan and README updates |
| 2 | `feat(prisma+lib):` | `Session` model + migration, `lib/session.ts`, `lib/request-meta.ts` |
| 3 | `feat(lib):` | Rewrite `auth.ts` + `cookies.ts`; strip token issuance from `oauth.ts` |
| 4 | `refactor:` | Codemod — strip `withAuth()` from 80 files |
| 5 | `feat(routes+collab):` | Wire login/OAuth/logout/collab/linking to sessions |
| 6 | `test:` | Rewrite affected tests; add `session.test.ts` |
| 7 | `chore:` | Drop `jose`, `JWT_SECRET` |

## Test plan

- [x] CI green: `test.yml`, `build-check.yml`, `migration-check.yml`, `codeql.yml`
- [ ] Burn in on dev Neon branch ~24h with no auth-error spikes
- [ ] Manual smoke: Google member login → land on `/hiring/reviewer`
- [ ] Manual smoke: Google partner (`@dartmouth.edu`) login → land on `/portal`
- [ ] Manual smoke: CAS login → land on `/portal`
- [ ] Manual smoke: `/dev-login-as?daliEmail=...` still works
- [ ] Manual smoke: Wait 30+ minutes idle, refresh → still logged in (no "bounced every 15 min" regression)
- [ ] Manual smoke: Logout → next request 302s to `/login`
- [ ] Manual smoke: Revoke a `Session` row in Prisma Studio → next request 302s to `/login`
- [ ] Collab smoke: open a Tiptap document (reviewer application or interviewer interview), presence indicator appears, content persists, two-tab cursor sync works
- [ ] Promote to staging; verify migration drops `RefreshToken` cleanly against prod snapshot
- [ ] Post `#dali-eng` heads-up ~24h before prod cut
- [ ] Promote to prod outside any hiring-cycle close window

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,17 +1,123 @@
-# DALI-OS
+# DALI OS
 
+Internal operations platform for the DALI Lab. Single React Router 7 app (`dali-api/`) that serves the UI, the JSON API, and the Hocuspocus realtime collab server. Originally a hiring tool, now expanding to projects, members, partners, calendar, and education.
 
+## Stack
 
+| Layer | Tech |
+|---|---|
+| App | React Router 7 (full-stack), React 19, TypeScript |
+| DB | Postgres 16 on Neon, Prisma 7 (`@prisma/adapter-neon` / `@prisma/adapter-pg`) |
+| Realtime | Hocuspocus + Yjs + Tiptap, Redis for fan-out |
+| Auth | Google OAuth, Dartmouth CAS, opaque session cookies + Bearer headers |
+| Styling | Tailwind CSS 4 |
+| Runtime | Node 22, npm |
+| Tests | Vitest (unit), Playwright (e2e) |
+| Deploy | Fly.io, branches `dev` → `staging` → `prod` |
 
-## Quickstart with Docker Compose
-| Service | Image | Ports | Description |
-| -------- | -------- | -------- | ----------- |
-| db | postgres:16-alpine | 5432 | PostgreSQL database |
-| api | ./dali-api | 3001, 5555 | Runs Prisma migrate, API, and Prisma Studio |
+## Repo layout
 
-## Environment Variables (with Docker Compose)
-| Variable | Description | Default Value |
-| -------- | ----------- | ------------- |
-| GOOGLE_CLIENT_ID | Google OAuth ID | None | 
-| GOOGLE_CLIENT_SECRET | Google OAuth Secret | None |
-- When using Docker Compose locally, only the Google OAuth credentials (in the top level .env) are required
+```
+dali-api/
+  app/
+    hiring/          cycles, reviewers, interviews, delibs, decisions
+    admin-console/   members, domains, role assignment
+    projects/        project list, staffing
+    members/         directory, user profiles
+    partners/        partner portal
+    calendar/        scheduling
+    collab/          Hocuspocus server + Yjs persistence
+    routes/          shared routes (auth, portal, oauth, uploads)
+    components/      shared UI
+    lib/             db client, auth, server utilities
+  prisma/            schema.prisma, migrations/, seed.ts
+  e2e/               Playwright specs
+.github/workflows/   CI/CD
+docker-compose.yml   local Postgres + API + Prisma Studio
+```
+
+## Local dev
+
+Prereqs: Docker, Node 22, a repo-root `.env` containing `GOOGLE_CLIENT_ID` and `GOOGLE_CLIENT_SECRET`.
+
+```bash
+docker compose up
+```
+
+Brings up Postgres, runs `prisma db push --force-reset && prisma db seed`, then starts the dev server on `:3001`, the collab server on `:3002`, and Prisma Studio on `:5555`.
+
+Bare metal (your own Postgres):
+
+```bash
+cd dali-api
+npm install
+npm run db:reset:local   # destroys local DB
+npm run dev
+```
+
+Skip login during dev at `/dev-login` (dev-only route, gated to non-prod builds).
+
+## Commands
+
+Run from `dali-api/`.
+
+| Task | Command |
+|---|---|
+| Unit tests | `npm test` |
+| E2E tests | `npm run test:e2e` (needs a seeded Postgres) |
+| Typecheck | `npm run typecheck` |
+| Build | `npm run build` |
+| Dev server | `npm run dev` |
+| Reset local DB | `npm run db:reset:local` |
+
+No ESLint/Prettier is wired up.
+
+## Environments
+
+| Env | Branch | Fly app | Neon branch | DB on deploy |
+|---|---|---|---|---|
+| Dev | `dev` | `dali-api-dev` | `development` | wipe → migrate → seed |
+| Staging | `staging` | `dali-api-staging` | `staging` | restore from prod → migrate |
+| Prod | `prod` | `dali-api-prod` | `production` | migrate only |
+
+Promotion is staged: PRs merge to `dev`, then `promote-to-staging.yml` and `promote-to-prod.yml` move code forward. Per-PR previews spin up their own Neon branch + Fly app via `preview-deploy.yml` and tear down on close.
+
+## Database & migrations
+
+Authoring rules:
+
+- Edit `prisma/schema.prisma`, then `npx prisma migrate dev --name <change>`.
+- Commit schema and migration in the same PR.
+- Never edit or delete an applied migration. Fix forward with a new one.
+- Flag data-losing migrations (drops, non-null without default) in the PR description.
+
+Runtime uses the pooled `DATABASE_URL`. `prisma migrate` needs a non-pooled URL (advisory locks don't survive PgBouncer transaction pooling); `prisma.config.ts` auto-derives it from `DATABASE_URL` or honors an explicit `DIRECT_URL`. Full detail: `dali-api/prisma/MIGRATIONS.md`.
+
+## CI gates
+
+Failures on these block merge:
+
+- `test.yml` — Vitest + Playwright against a real Postgres service container.
+- `build-check.yml` — Docker build via flyctl.
+- `migration-check.yml` — schema/migration drift, deleted-migration guard, pgfence safety analysis.
+- `codeql.yml` — static security scan.
+- `preview-deploy.yml` — per-PR Neon + Fly preview (only blocks if it fails specifically).
+
+## Auth surface
+
+- `/login` — Google OAuth or Dartmouth CAS.
+- `/portal/*` — applicant portal (lighter layout, no internal nav).
+- `/oauth/*` — DALI OS acts as an OAuth provider for the `dali-os-mcp` integration (`authorize`, `token`, `revoke`, `callback/*`).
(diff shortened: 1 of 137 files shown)


Kiran-Jones:
**test-api fix (ce32a5e)**

`app/lib/__tests__/cookies.test.ts` was the only failing suite — 1 failed suite, 682/682 actual test cases still passing. Root cause:

```
Cannot find module '../generated/prisma/client.js' imported from .../lib/db.ts
 ❯ app/lib/db.ts:1:1
 ❯ app/lib/session.ts:2:1
```

The import chain was `cookies.test.ts` → `~/lib/cookies` → `ROLLING_TTL_MS` from `~/lib/session` → `prisma` from `~/lib/db` → `../generated/prisma/client.js`. CI's `test-api` job runs `npm test` without `prisma generate`, so the generated client wasn't present at module-load time.

Other test files mock `~/lib/db` (or don't touch it transitively), so they were fine. `cookies.test.ts` only imports cookie helpers and never calls any Prisma function — but the constant import was enough to trip the loader.

**Fix:** inlined a local `SESSION_COOKIE_MAX_AGE_SECONDS = 30 * 24 * 60 * 60` in `cookies.ts` so it no longer imports from `session.ts`. The cookie Max-Age and the DB rolling TTL were always independently enforced anyway (server is source of truth — an expired DB session 302s on next request regardless of cookie lifetime), so they don't need to share a constant.

Verified locally by deleting `app/generated/` and running `cookies.test` in isolation against the fresh state — 10/10 pass.

Not touching `.github/workflows/test.yml` per CLAUDE.md scope rules; the breakage came from the application code making a new assumption about what CI loads, and the application-code fix is cleaner anyway.

sophp516 (approved):
LGTM

PR #520, opened 2026-05-14
Always enforce Google domain + drop account-type cookie + OAuthSession enums (issues E, H)

Kiran-Jones:
Closes / addresses Known auth-surface issues **E** and **H** in `dali-os-mcp.md`.

## Summary

**Issue E - always-enforce `@dali.dartmouth.edu` on `/auth/callback/google`.**
The domain check was previously gated on the `__dali_account_type=member`
cookie. If the cookie was missing or stripped, the check fell open and the
route's email-domain branches silently provisioned a Dartmouth-student or
partner User for someone who had clicked the Member button. The Member
button is the only `/login` path that uses this callback, so
`@dali.dartmouth.edu` is the only valid outcome. The check is now
unconditional, the `__dali_account_type` cookie is gone (the action no
longer sets it; the callback no longer reads or clears it), and the
now-unreachable dartmouth/partner branches in `/auth/callback/google` have
been removed. The OAuth-provider `/oauth/callback/google` still handles
those branches but gates on `OAuthSession.accountType`, which is
server-stored.

**Issue H - add Prisma enums for `OAuthSession`.** `provider` and
`accountType` were free-form strings, validated by route logic only.
Convert them to Prisma enums:

```prisma
enum OAuthProvider     { google, cas }
enum OAuthAccountType  { member, dartmouth, partner }
```

The migration is a 1:1 USING text->enum relabel (all existing rows already
match the new labels). `createOAuthSession` is now typed against the
enums, and `/oauth/authorize` validates `account_type` query values up
front rather than letting a bad value fail at INSERT.

## User-visible behavior change

The Member-button login flow used to silently log a user in as a
Dartmouth student or partner if the `__dali_account_type` cookie was
absent. After this change, any non-`@dali.dartmouth.edu` email at
`/auth/callback/google` returns to `/login?error=access_denied` with the
existing "Only @dali.dartmouth.edu accounts are allowed" message. The
Applicant (CAS) flow is unchanged. The Google `hd=dali.dartmouth.edu`
hint on Google's account picker is still set (it's a UX nudge, not a
security boundary).

## Test plan

- [ ] CI green (unit, e2e, migration-check, build-check, codeql)
- [ ] Manual smoke: Member-button login with `@dali.dartmouth.edu` -> success
- [ ] Manual smoke: Member-button login with a `@dartmouth.edu` student account -> `access_denied`
- [ ] Manual smoke: Applicant CAS login still works
- [ ] Migration runs cleanly on the staging Neon branch via the standard deploy path

## Concurrent PR notice

PR `claude/cleanup-calendar-link-route` (calendar-link split, issue F)
is running concurrently and touches a different section of
`auth.callback.google.ts` (the `cal:` state dispatch at lines ~37-61, untouched here).
Mergeable in either order; second-in gets a small rebase.

### DIFF
--- a/dali-api/app/lib/oauth.ts
+++ b/dali-api/app/lib/oauth.ts
@@ -1,6 +1,7 @@
 import { createHash, randomBytes } from "node:crypto";
 import { OAuth2Client } from "google-auth-library";
 import { prisma } from "~/lib/db";
+import type { OAuthProvider, OAuthAccountType } from "~/generated/prisma/enums";
 
 // types
 
@@ -97,8 +98,8 @@ export async function createOAuthSession(params: {
   codeChallengeMethod: string;
   redirectUri: string;
   state: string;
-  provider: string;
-  accountType?: string;
+  provider: OAuthProvider;
+  accountType?: OAuthAccountType;
 }) {
   return prisma.oAuthSession.create({
     data: {
--- a/dali-api/app/routes/__tests__/login.test.ts
+++ b/dali-api/app/routes/__tests__/login.test.ts
@@ -11,7 +11,6 @@ import { action } from "~/routes/login";
 function makeRequest(ip = "1.2.3.4") {
   const form = new URLSearchParams({
     provider: "google",
-    accountType: "member",
   });
   return new Request("http://localhost/login", {
     method: "POST",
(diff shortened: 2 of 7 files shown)


Kiran-Jones:
Rebased onto current dev (PR #518 merged). One conflict in `auth.callback.google.ts` body — the PR originally deleted the inline dartmouth/partner branches (because they were unreachable after always-enforce), but #518 had since moved them into `lib/user-provisioning.ts`. Resolved by keeping the always-enforce check (this PR's intent) and using the helper call from #518 — the helper's member branch fires correctly, the unreachable dartmouth/partner branches stay in the helper for the OAuth-provider callback. Typecheck clean, all 703 tests passing locally. Force-pushed; CI re-running.

PR #531, opened 2026-05-14
v0 Phase 2 — destructive identity refactor (DALIMember slim-down + hi…

Kiran-Jones:
…ring FK renames) (#525)

* v0 Phase 2 — destructive identity refactor (DALIMember slim-down + hiring FK renames)

Schema:
- User: drop googleAccessToken/RefreshToken/TokenExpiresAt (moved to GmailIntegration).
- DALIMember: slim to thin marker — drop firstName/lastName/daliEmail/ dartmouthEmail/did/roles. userId NOT NULL. did data backfills into User.netId.
- Hiring FKs (9 tables) now point at User instead of DALIMember: CycleReviewer.userId, CycleInterviewer.userId, DomainLeadAssignment.userId (renamed from daliMemberId/memberId, plus DomainLeadAssignment.termId NOT NULL), plus Decision.madeById, ApplicationReview.submittedById, DelibsSession.openedById, LegacyEmailTemplate.createdById, EmailTemplateVersion.createdById, ConfidentialityAgreementVersion.createdById (column names unchanged; FK target swapped to User).
- Domain: code + displayName tightened to NOT NULL.
- OAuthAccountType enum collapsed to `member` only.
- MemberRole enum dropped (Admin → AdminMembership rows, HiringLead → CoreAssignment with leadTitle="Hiring Lead").

Migration (20260514130959_v0_phase2_identity_refactor):
- Single transaction. Designed for a maintenance window with zero active write traffic. Expected runtime < 5s on <10k rows. Halts with a clear error if no Term row exists (run npm run db:seed:v0-reference first).
- Backfills run before drops so no data is lost: 116 orphan DALIMember rows promote to User rows; MemberRole[] values translate to AdminMembership + CoreAssignment; hiring FK columns get userId-translated values from the DALIMember.userId join; GmailIntegration absorbs the legacy User.google* tokens; DomainLeadAssignment.termId backfills to current term.

Code:
- lib/roles.ts rewritten. UserRoles shape: { isLabMember, isHiringLead, isAdmin, isDomainLead }. New helpers: currentTerm(), requireMember().
- lib/user-provisioning.ts: upsertUserFromGoogle collapsed to @dali.dartmouth.edu branch only — non-DALI calls throw. Partner/dartmouth Google branches were unreachable in prod (the website only offers Google for members; /oauth/* is MCP-only and MCP is members-only).
- lib/collabAuth.ts: ownership checks compare to auth.user.sub directly instead of indirecting through DALIMember.id. Also fixed a Phase-1 bug where the DALIMember-id-vs-User-id mismatch broke the owner check.
- lib/google-calendar.ts: legacy fetchBusyFromLegacyUserTokens removed. Free-busy is now exclusively via UserCalendarLink.
- lib/gmail-integration.ts: new helper module (getApplicationsGmailRefreshToken, isApplicationsGmailConnected). Six Gmail-send call sites switched over.
- ~30 hiring routes: prisma.dALIMember.findFirst({where:{userId}}) patterns replaced with direct User reads via auth.user.sub. daliMember relation includes/access rewritten to use the user relation directly.
- Admin Console members/domains UIs rewritten to manage AdminMembership / CoreAssignment / DomainLeadAssignment rows. The MemberRole[] editor is replaced with toggles backed by row insert/delete.
- /oauth/authorize: VALID_ACCOUNT_TYPES collapsed to ["member"].
- All tests updated to match the new schema/code shape. 698/698 passing.

Coordinated with V0_PHASE2_PLAN.md. Deploy expects:
1. 2026-S cycle status = Completed.
2. Neon snapshot taken immediately before the Fly release.
3. Zero active write traffic during the ~1-minute pod-swap window.

* fix(migration): make Phase 2 tolerant of empty Term table for CI

Two CI failures (Check migrations + e2e-api) both root-caused to my Phase 2 migration raising an exception when no Term row exists. CI applies migrations against a fresh DB before any seed runs, so no Terms are present — the preemptive RAISE blocked the migration from completing.

Fix:
- Only raise when there's actual data that needs a Term reference (HiringLead members to translate, OR DomainLeadAssignment rows with NULL termId). With an empty DB, the DO block now returns cleanly.
- Drop the explicit BEGIN/COMMIT wrapper. Prisma already runs each migration in its own transaction; the redundant pair was making the "current transaction is aborted" error surface in CI when the inner RAISE fired, instead of letting Prisma's transaction handling roll back cleanly.

The prod safety net is still strong:
- If HiringLead members or NULL-termId DLAs exist but no Term, the RAISE fires with a clear actionable message.
- If somehow that's bypassed, the subsequent `ALTER COLUMN termId SET NOT NULL` would fail with a postgres constraint error and roll back.

* fix(seed): seed a Term before the test hiring-lead row references it

Phase 2 made DomainLeadAssignment.termId NOT NULL and CoreAssignment.termId required. The existing local seed looked up "any Term" via findFirst and fell back to undefined when none existed — which is exactly the case in CI (`prisma migrate reset --force && prisma db seed` runs against a fresh DB with no Terms yet).

Upsert the "26S" term at the top of seed.ts so the downstream CoreAssignment and DomainLeadAssignment writes have a real termId to point at. Prod uses prisma/seeds/v0-reference.ts which seeds a full 12-term window; locally one term is enough for the hiring-lead + domain-lead fixtures.

* fix(seed): align local Domain.displayName with legacy name for e2e parity

Phase 2's admin-console.domains UI renders domain.displayName instead of domain.name. The local seed had set displayName to the canonical catalog labels ("UI/UX Design", "Fullstack Dev", "Product Management"), breaking e2e tests that match the rows on "Design"/"Engineering"/"Product".

Mirror displayName to name in the local seed only. Prod still uses the canonical catalog labels via prisma/seeds/v0-reference.ts.

* fix: pass User.id (not DALIMember.id) where hiring FKs expect userId

After my Phase 2 perl bulk-replace renamed daliMemberId → userId across the hiring routes, several call sites were still passing `member.id` as the value — which is the DALIMember.id (cuid for the marker row), not the User.id the renamed columns now expect. The reviewer dashboard, my- availability, my-interviews, interview-complete, and interviewer routes all silently returned 0 rows because the where-clauses didn't match.

Surface caught by the reviewer.spec.ts e2e failure ("Assigned Written Applications" heading never appeared because cycleReviewer lookup found 0 matching rows when filtered on userId = DALIMember.id).

Files touched:
- reviewer.tsx — pass auth.user.sub to inferUnderReviewStage instead of member.id (which the function uses as a userId filter on CycleInterviewer).
- cycles.ts — rename inferUnderReviewStage's `memberId` parameter to `userId` to make the intent explicit.
- api.cycles.$cycleId.my-availability.ts, my-interviews.$interviewId.notes.ts — perl had injected `auth.user.sub` into helper functions where `auth` isn't in scope; fix to use the helper's `userId` parameter.
- api.decisions.lineage.test.ts — Phase 2 stores auth.user.sub in Decision.madeById; update the assertion to match.

* chore: add Phase 1 verification SQL script

Run against staging (or any post-Phase-1 DB) to confirm:
- All 43 new tables present
- New User columns added (all nullable)
- Legacy columns intact (Phase 1 is additive)
- Existing data unchanged (User, DALIMember, hiring tables)
- v0-reference seed status — flags whether Term needs seeding before Phase 2 deploy

Usage:
  flyctl proxy 5432:5432 -a dali-api-staging & psql "$STAGING_URL" -f scripts/verify-v0-phase1.sql

or paste into Neon's SQL editor.

* fix(seed): heal legacy Domain rows on v0-reference seed

When the v0-reference seed ran against staging, it upserted by `code` — which legacy prod Domain rows didn't have. Result: 12 v0-reference rows created alongside 12 legacy rows, with all live cycle data (challenge versions, application cycles, lead assignments, reviewers, interviewers) still attached to the legacy rows.

Phase 2's `Domain.code SET NOT NULL` + unique constraint would have failed on the 6 name-collision cases (Engineering, Animation, Graphics, Writing, Videography, Photography).

Make the seed self-healing: each DomainSeed now lists its `legacyNames`, and the seed checks for a code-NULL row matching any of them. If found:
1. Delete the empty v0-reference duplicate (no FK references — safe).
2. Update the legacy row in place with the canonical code/displayName.

This preserves all live cycle data and runs idempotently. Re-running the seed against staging cleans up the dirty state. When the seed runs against prod for the first time, it heals legacy rows in-place — no duplicates created.

* fix(seed): use raw SQL for code IS NULL filter in heal step

Prisma 7's generated StringFilter for nullable columns doesn't include null in the type union, and the runtime ALSO rejects {code: null} on findFirst with a PrismaClientValidationError ("Argument code must not be null"). Casting to any silenced the type error but the runtime still threw.

Raw $queryRaw with IS NULL is clearer and works against both the typing and runtime constraints.

---------

### DIFF
--- a/dali-api/app/admin-console/__tests__/api.domains.create.test.ts
+++ b/dali-api/app/admin-console/__tests__/api.domains.create.test.ts
@@ -48,7 +48,9 @@ describe("POST /api/domains — name validation", () => {
       context: {},
     } as any);
     expect(res.status).toBe(201);
-    expect(mockPrisma.domain.create).toHaveBeenCalledWith({ data: { name: "Design" } });
+    expect(mockPrisma.domain.create).toHaveBeenCalledWith({
+      data: { name: "Design", code: "Design", displayName: "Design" },
+    });
   });
 
   it("rejects an empty name", async () => {
--- a/dali-api/app/admin-console/__tests__/api.members.roles.test.ts
+++ b/dali-api/app/admin-console/__tests__/api.members.roles.test.ts
@@ -13,34 +13,67 @@ vi.mock("~/lib/cors", () => ({
 
 import { prisma } from "~/lib/db";
 import { requireAuth } from "~/lib/auth";
-import { isAdmin } from "~/lib/roles";
+import { isAdmin, currentTerm } from "~/lib/roles";
 import { action } from "~/admin-console/routes/api.members.$memberId.roles";
 
 const ADMIN_ID = "admin-1";
-const MEMBER_ID = "member-1";
+const USER_ID = "user-1";
+const TERM_ID = "term-1";
+
+// Phase 2: the endpoint translates legacy role labels ("Admin", "HiringLead")
+// into AdminMembership / CoreAssignment writes. `memberId` in the route param
+// is now the User.id.
 
 const mockPrisma = prisma as unknown as {
-  dALIMember: {
+  adminMembership: {
     findUnique: ReturnType<typeof vi.fn>;
-    update: ReturnType<typeof vi.fn>;
+    upsert: ReturnType<typeof vi.fn>;
+    deleteMany: ReturnType<typeof vi.fn>;
   };
+  coreAssignment: {
+    findFirst: ReturnType<typeof vi.fn>;
+    create: ReturnType<typeof vi.fn>;
+    deleteMany: ReturnType<typeof vi.fn>;
+  };
+  user: {
+    findUnique: ReturnType<typeof vi.fn>;
+  };
+  $transaction: ReturnType<typeof vi.fn>;
 };
 
 beforeEach(() => {
   vi.clearAllMocks();
-  (mockPrisma as any).dALIMember = {
-    findUnique: vi.fn().mockResolvedValue({ roles: [] }),
-    update: vi.fn().mockResolvedValue({ id: MEMBER_ID, roles: ["Admin"] }),
+  (mockPrisma as any).adminMembership = {
+    findUnique: vi.fn().mockResolvedValue(null),
+    upsert: vi.fn(),
+    deleteMany: vi.fn(),
   };
+  (mockPrisma as any).coreAssignment = {
+    findFirst: vi.fn().mockResolvedValue(null),
+    create: vi.fn(),
+    deleteMany: vi.fn(),
+  };
+  (mockPrisma as any).user = {
+    findUnique: vi.fn().mockResolvedValue({
+      id: USER_ID,
+      firstName: "Ada",
+      lastName: "Lovelace",
+      daliEmail: "ada@dali.dartmouth.edu",
+      adminMembership: { id: "a1" },
+      coreAssignments: [],
+    }),
+  };
+  (mockPrisma as any).$transaction = vi.fn(async (cb: (tx: unknown) => Promise<unknown>) => cb(mockPrisma));
   vi.mocked(requireAuth).mockResolvedValue({
     ok: true,
     user: { sub: ADMIN_ID, email: "a@x.com", type: "user" },
   } as any);
   vi.mocked(isAdmin).mockResolvedValue(true);
+  vi.mocked(currentTerm).mockResolvedValue({ id: TERM_ID } as any);
 });
 
 function makeRequest(body: unknown) {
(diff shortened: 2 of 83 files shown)


sophp516 (approved):
LGTM

PR #533, opened 2026-05-14
v0 Phase 2 follow-up — migration collision handling + verify scripts

Kiran-Jones:
## Summary

Follow-up to merged #525. These commits were authored after #525 was merged, in the course of validating the migration against real prod-shaped data on staging. Two of them are **critical fixes** for the Phase 2 migration that surface on real data; without them, the next staging redeploy and the eventual prod deploy will both fail the same way today's staging deploy did.

### Migration fixes (must-merge before any further deploys)

- **`391696e` — netId unique collision handling.** Staging deploy hit:
  ```
  ERROR: duplicate key value violates unique constraint "User_netId_key"
  DETAIL: Key ("netId")=(f006jnh) already exists.
  ```
  Root cause: a member logged in via CAS first (User.netId='f006jnh'), and a separate DALIMember row linked to a different User had did='F006JNH'. M3a tried to set the second User's netId to LOWER('F006JNH')='f006jnh' — collision. Fix: M2 INSERT now uses `ON CONFLICT DO NOTHING` (bare) so any unique violation skips cleanly, and computes netId as NULL if some other User already owns the target value; M3a UPDATE adds a `NOT EXISTS` guard for the same scenario. The second User can complete their identity later via CAS sign-in (lib/auth/linking.ts handles the merge).

- **`d0a63a5` — FK drop ordering.** Second staging failure:
  ```
  ERROR: insert or update on table "ApplicationReview" violates foreign key constraint "ApplicationReview_submittedById_fkey"
  DETAIL: Key (submittedById)=(cmox...) is not present in table "DALIMember".
  ```
  M3f translates six "*ById" columns from DALIMember.id values to User.id values. Postgres validates the existing FK on every UPDATE — and those FKs still pointed at DALIMember. Fix: move the FK drops into a new M3-pre block before the translations.

### Seed self-healing (so v0-reference doesn't create domain duplicates)

- **`0249153` + `c56d16e` — `prisma/seeds/v0-reference.ts` heals legacy Domain rows in place** instead of creating duplicates alongside them. Prod had 12 legacy Domain rows with `code IS NULL`; without the heal, Phase 2's `Domain.code SET NOT NULL` would fail. The seed is now idempotent against any starting state (fresh DB, pre-Phase-1 prod, post-Phase-1-with-dupes staging).

### Verification scripts (nice-to-have, used during staging validation)

- **`04f6549`** — `scripts/verify-v0-phase1.sql` — 12-section schema + data integrity check
- **`7e93d3e`** — `scripts/verify-v0-phase2.sql` — analogous Phase 2 checks (FK retargeting, NOT NULL tightening, backfill correctness, sample joins)

## Validation status

The migration in this PR has been **applied successfully against staging Neon today** (which mirrors prod via the rebuild-from-snapshot mechanism). All 16 verification checks pass:

| Check | Result |
|---|---|
| DALIMember slimmed to 4 cols | ✓ |
| User.google* dropped | ✓ |
| Hiring FKs renamed to userId | ✓ |
| NOT NULL constraints tightened | ✓ |
| MemberRole + OAuthAccountType enum cleanup | ✓ |
| All 10 hiring FKs now reference User | ✓ |
| 0 orphan DALIMembers (95 → User rows) | ✓ |
| User count 215 → 311 | ✓ |
| Hiring data preserved (45/18/118/0/0/42) | ✓ |
| AdminMembership × 9 + Hiring Lead CoreAssignment × 8 (matches Phase 1 roles[] counts) | ✓ |
| GmailIntegration migrated for applications@dali.dartmouth.edu | ✓ |
| Sample reviewer join through new FK shape | ✓ |

## Test plan

- [x] `npm run typecheck` — clean
- [x] `npm test` — 698/698 passing
- [x] Migration applied against staging Neon successfully (after 2 instructive failures that the 2 migration fixes address)
- [x] All 16 `verify-v0-phase2.sql` checks green on staging
- [ ] Merge to `dev` to put dev in sync with what staging currently has
- [ ] Next staging redeploy should apply cleanly on first try (no manual recovery needed)
- [ ] Then promote to prod during a maintenance window

### DIFF
--- a/dali-api/app/admin-console/__tests__/api.domains.create.test.ts
+++ b/dali-api/app/admin-console/__tests__/api.domains.create.test.ts
@@ -48,7 +48,9 @@ describe("POST /api/domains — name validation", () => {
       context: {},
     } as any);
     expect(res.status).toBe(201);
-    expect(mockPrisma.domain.create).toHaveBeenCalledWith({ data: { name: "Design" } });
+    expect(mockPrisma.domain.create).toHaveBeenCalledWith({
+      data: { name: "Design", code: "Design", displayName: "Design" },
+    });
   });
 
   it("rejects an empty name", async () => {
--- a/dali-api/app/admin-console/__tests__/api.members.roles.test.ts
+++ b/dali-api/app/admin-console/__tests__/api.members.roles.test.ts
@@ -13,34 +13,67 @@ vi.mock("~/lib/cors", () => ({
 
 import { prisma } from "~/lib/db";
 import { requireAuth } from "~/lib/auth";
-import { isAdmin } from "~/lib/roles";
+import { isAdmin, currentTerm } from "~/lib/roles";
 import { action } from "~/admin-console/routes/api.members.$memberId.roles";
 
 const ADMIN_ID = "admin-1";
-const MEMBER_ID = "member-1";
+const USER_ID = "user-1";
+const TERM_ID = "term-1";
+
+// Phase 2: the endpoint translates legacy role labels ("Admin", "HiringLead")
+// into AdminMembership / CoreAssignment writes. `memberId` in the route param
+// is now the User.id.
 
 const mockPrisma = prisma as unknown as {
-  dALIMember: {
+  adminMembership: {
     findUnique: ReturnType<typeof vi.fn>;
-    update: ReturnType<typeof vi.fn>;
+    upsert: ReturnType<typeof vi.fn>;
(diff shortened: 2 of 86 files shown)


Kiran-Jones:
Misleading diff — branch never picked up the squash-merge of #525, so GitHub shows the full Phase 2 history. Re-opening with a clean branch off current dev that contains only the actual missing changes (2 files).

PR #535, opened 2026-05-14
v0 Phase 2 follow-up — migration collision handling + verify script (…

Kiran-Jones:
…#534)

Two fixes to the Phase 2 migration discovered during staging deploy validation against real prod-cloned data. Without these, the next staging redeploy and any prod deploy will both fail with the same errors today's staging deploy did.

1. netId unique collision (M2 + M3a) Staging failed with: ERROR: duplicate key value violates unique constraint "User_netId_key" DETAIL: Key ("netId")=(f006jnh) already exists.

   Root cause: a member logged into the app via CAS first (User.netId was set), and a separate DALIMember row linked to a different User had did='F006JNH'. The migration tried to set the second User's netId to LOWER('F006JNH') — collision.

   Fix:
   - M2 INSERT computes netId as NULL when some other User already owns the target value, and widens ON CONFLICT to bare DO NOTHING so any unique violation skips the row.
   - M3a UPDATE adds a NOT EXISTS guard that skips the write when it would collide. The orphaned User can complete their identity via CAS sign-in later — lib/auth/linking.ts:linkCasToGoogleUser merges the two User rows when that happens.

2. FK drop ordering (new M3-pre block) Staging then failed with: ERROR: insert or update on table "ApplicationReview" violates FK "ApplicationReview_submittedById_fkey" DETAIL: Key (submittedById)=(cmox...) is not present in DALIMember.

   M3f translates six "*ById" columns from DALIMember.id values to User.id values. Postgres validates the existing FK on every UPDATE, and those FKs still pointed at DALIMember.

   Fix: move the nine FK drops (for the tables whose values get translated) into a new M3-pre block before M3e/M3f run. Two FK drops that don't involve value translation (DALIMember_userId_fkey, DomainLeadAssignment_termId_fkey) stay in M4.

3. New scripts/verify-v0-phase2.sql Mirrors verify-v0-phase1.sql for the destructive refactor. 16 sections covering schema (DALIMember slim-down, User.google* drop, FK renames, NOT NULL tightening, MemberRole/OAuthAccountType enum cleanup) plus data integrity (no orphans, no NULL FKs, hiring counts preserved, AdminMembership/CoreAssignment/GmailIntegration backfilled).

Validated against staging Neon today — all 16 verification sections green. Staging Neon is a fresh prod snapshot per deploy, so this is effectively a dry-run of the prod deploy.

### DIFF
--- a/dali-api/prisma/migrations/20260514130959_v0_phase2_identity_refactor/migration.sql
+++ b/dali-api/prisma/migrations/20260514130959_v0_phase2_identity_refactor/migration.sql
@@ -31,6 +31,10 @@ ALTER TABLE "DomainLeadAssignment" ADD COLUMN "userId_new" TEXT;
 -- We reuse the DALIMember.id as the new User.id (both cuids; no collision
 -- since DALIMember.id is unique and no existing User shares it).
 
+-- INSERT uses ON CONFLICT DO NOTHING (without column spec) so that
+-- conflicts on ANY unique constraint (daliEmail, netId, or did's lowercase
+-- collision with a pre-existing User.netId) skip cleanly. Skipped rows are
+-- handled by the linking UPDATE below, which keys on daliEmail.
 INSERT INTO "User" (id, "createdAt", "updatedAt", "firstName", "lastName", "daliEmail", "netId")
 SELECT
   m.id,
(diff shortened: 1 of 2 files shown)


sophp516 (approved):
LGTM

PR #538, opened 2026-05-14
DALI MCP v1 — OAuth foundation + /mcp whoami

Kiran-Jones:
## Summary

Wires up the MCP v1 surface end-to-end:
- **OAuthClient registry** replaces the hardcoded `VALID_CLIENT_IDS` array. `/oauth/authorize` and `exchangeAuthorizationCode` look the client up in Postgres; `claude-desktop` and `claude-code` are seeded (no `dali-api` seed per `schema.prisma:2206`).
- **Per-client redirect-URI matching** with RFC 8252 loopback semantics for native clients (any port on 127.0.0.1/localhost over http, registered path required; rejects https loopback, 0.0.0.0, public IPs).
- **Per-client provider + accountType + scope enforcement** at the authorize boundary. MCP clients are pinned to Google + `member` + `requireMembership: true`.
- **Existing-session shortcut** at `/oauth/authorize` — a valid cookie session skips the Google round-trip and branches straight to consent (or direct code-issue if a matching grant exists).
- **Consent screen** (`/oauth/consent`) with cookie-session ownership check; approve writes the `OAuthGrant`, deny redirects back with `error=access_denied`.
- **`OAuthGrant` writeback** at `/oauth/token`: upserts (scopes are union-merged across re-auths), links the issued `Session.grantId`, bumps `lastUsedAt`.
- **`.well-known/oauth-authorization-server`** — RFC 8414 metadata.
- **`/settings/connected-apps`** — list + revoke; revoke cascades via `revokeAllForGrant`.
- **`lib/mcp-auth.ts`** — resolves Bearer → Session → Grant → Client, re-runs `requireMembership` on every request (off-boarding kill switch), bumps `lastUsedAt` + rolls session.
- **`/mcp`** — hand-rolled JSON-RPC 2.0 over HTTP. Implements `initialize`, `tools/list`, `tools/call`, `ping`, and the `notifications/*` no-ops. Per-grant rate limit at 120 req/min keyed `grant:<id>`.
- **`whoami`** tool — single tool in v1; returns `{ id, daliEmail, netId, firstName, lastName, tier }`. Requires `mcp:read`.
- **Audit log**: every tool call writes `mcp.tool_called` with `{ toolName, clientId, clientName, grantId }`.
- **`/help/mcp`** — member docs for Claude Code / Claude Desktop setup.

## Migrations (both additive)

1. `20260514150000_oauth_session_scopes_client_id` — adds `OAuthSession.clientId String?` and `OAuthSession.scopes String[] @default([])`.
2. `20260514150100_session_grant_id_fk` — adds the FK constraint `Session.grantId → OAuthGrant.id` (Prisma relation was already declared; this lands the DB-level constraint).

No data-loss. Both should pass pgfence cleanly.

## SDK-less choice

The plan suggested `@modelcontextprotocol/sdk` with a hand-roll fallback. For a single-tool v1 the hand-rolled JSON-RPC handler is ~140 lines and avoids both a new dependency and the fetch↔node-http adapter cost. If a future phase adds streaming/SSE we can revisit.

## Test plan

- [x] Unit: `npm test` — 701 tests pass (added redirect-URI matching coverage; refactored existing oauth tests around the new `getOAuthClient` shape).
- [x] Typecheck: `npm run typecheck` clean.
- [ ] E2E: `e2e/mcp-auth-flow.spec.ts` walks discovery → authorize (existing-session shortcut) → consent → token → `/mcp whoami` and a second test confirms a plain cookie session is rejected at `/mcp`. Couldn't run locally — Docker postgres is in an unhealthy state on this machine. CI will exercise this.

## Files for reviewer attention

- Route filename: I used `routes/well-known.oauth-authorization-server.ts` and registered it explicitly in `app/routes.ts` to dodge React Router 7's dot-prefix quirk.
- `/oauth/consent` uses a `.tsx` because it renders. Help + settings pages do the same.
- Off-boarding kill switch lives in `mcp-auth.ts` (re-checks `DALIMember` per request).


Kiran-Jones:
Pushed a fix to drop the redundant `20260514150100_session_grant_id_fk` migration. The `Session_grantId_fkey` FK was already added by `20260514040346_v0_phase1_additive` when `OAuthGrant` originally landed — the plan's verified-facts comment was stale. The deleted migration was branch-only (never applied to dev), so `migration-check` 's no-deleted-applied-migration rule is unaffected.

This should clear `Check migrations`, `e2e-api`, and `Deploy API` together, since all three failed on the same `constraint already exists` error from that release_command/migrate step.

Kiran-Jones:
Pushed a follow-up: switched the consent redirect (and the authorize fallback-error redirect) to relative paths. The full-stack dali-api app serves both API + UI from one origin, so `FRONTEND_URL` (`http://localhost:5173`) pointed at a port that isn't bound in e2e/CI — hence the `ERR_CONNECTION_REFUSED` from `page.goto`.

PR #540, opened 2026-05-14
Intern → Full-time application cycles

Kiran-Jones:
## Summary
Adds a new `cycleType=InternToFull` alongside the existing Standard hire cycle so current-term interns (ERAS / EEJUST / WISP) can apply to convert to full-time members through a shortform application — no challenges, no interviews, single delibs round, and acceptance grants P1 `DomainEligibility` in the target domain.

The applicant flow lives **inside** the authenticated app (`/intern-to-full`) rather than the CAS-authed `/portal` so interns use their existing DALI session and don't end up creating duplicate user identities.

## What's in here

**Schema (additive)** — `prisma/migrations/20260514180412_intern_to_full_cycles/`
- New `ApplicationCycleType` enum + `cycleType` discriminator on `ApplicationCycle`
- New `InternToFullFormVersion` table (immutable shortform, mirrors `ChallengeVersion`)
- `Application.generalChallengeVersionId` made nullable; new `internToFullFormVersionId`
- `DomainApplication.challengeVersionId` made nullable; new direct `domainId` FK (InternToFull DomainApps link Domain directly since there's no challenge)

**Eligibility** — `app/hiring/lib/intern-eligibility.ts`
- `isInternToFullEligible(userId)` joins `ProjectAssignment` × active `Term` × `Domain.isInternProgram=true`. Fails closed when between terms (no fallback to the upcoming term).

**Concurrent active cycles** — `app/hiring/lib/cycles.ts`
- `getActiveCycle(cycleType?)` / `findOtherActiveCycleId(id, cycleType?)` scope to one cycleType. A Standard hire cycle and an InternToFull cycle may be Open concurrently; one-per-type still enforced.

**Applicant portal** — `app/routes/intern-to-full.tsx`
- Authenticated-member route, gated by the eligibility helper. Renders the shortform, target-domain multi-select, save-draft / submit / withdraw. Server-side validation on submit.

**Lead cycle setup** — `app/hiring/routes/lead.intern-to-full-cycle.$id.tsx`
- Lightweight setup page: pick/create `InternToFullFormVersion`, pick target domains, per-domain rubric, assign reviewers (min 2). `lead.cycle.$id` redirects here when `cycleType=InternToFull` so leads don't see challenge/interview UI that doesn't apply.
- `lead.tsx` new-cycle modal exposes the cycleType picker.

**Reviewer + delibs**
- `reviewer.application.$id` and `domain-lead.application.$id` loaders are null-safe across challengeVersion vs direct-domain shapes, and load the right "general" question set (challenge vs shortform) based on cycleType.
- `api.cycles.$cycleId.delibs` rejects opening an `Initial` session for InternToFull cycles.
- `domain-lead.delibs.$id` accepts InternToFull DomainApps into the `Final` lane once all reviews are submitted (no `interviews.some Completed` gate).

**Decision release** — `api.decisions.$id.release.ts`
- Resolves the target domain via `da.domain` or `da.challengeVersion.domain` (whichever is set). Fails closed if neither.
- On Released + Accepted + cycleType=InternToFull, upserts `DomainEligibility{level: P1, promotedBy: lead}` on the (userId, domainId) unique constraint. Intern eligibility is left intact and lapses naturally with the term.

**Notifications**
- Reuses the existing `CycleDecisionEmail` plumbing — each InternToFull cycle gets its own per-DecisionType bindings, so leads write conversion-specific copy per cycle. No new `DecisionType` or `NotificationType` enum values.

**One-line scope additions outside the feature**
- Tests across `cycles`, `decisions release`, `decisions lineage`, `reviewer.application` and `lead.cycle final-decisions-loader` updated to reflect the new mocks (cycleType filter, cycleType early-redirect lookup, new OR-filter shape on DomainApplication queries).

## Test plan
- [ ] **Local migrate**: `npm run db:reset:local && npx prisma migrate deploy` applies cleanly
- [ ] **Eligibility**: a user with a current-term ProjectAssignment in an `isInternProgram` domain can load `/intern-to-full`; a user without one gets the "Not eligible" view
- [ ] **Lead setup**: hiring lead creates an InternToFull cycle, picks 2 target domains, creates a new shortform with 2 questions, picks rubrics, assigns ≥2 reviewers per domain, sets close date, marks each domain ready
- [ ] **Open**: status route accepts the Draft→Open transition (with another Standard cycle also Open in parallel)
- [ ] **Apply**: intern fills shortform, picks domain(s), saves draft, submits; required-question validation fires on submit when blank
- [ ] **Review**: assigned reviewer opens the application; sees shortform answers; rubric scoring works
- [ ] **Delibs**: domain lead can open only a `Final` session (Initial returns 400); applications appear once reviews are submitted; close session drafts decisions
- [ ] **Release**: hiring lead releases an Accepted decision; verify a new `DomainEligibility{level: P1}` row is created for the applicant in the target domain
- [ ] **Withdraw**: applicant can withdraw post-submit and the route blocks edits


Kiran-Jones:
@claude how can I test this with the seed data?

Kiran-Jones:
Pushed `4b48e2d` addressing three review items from the InternToFull flow discussion:

**1. `DomainApplication.domainId` is now `NOT NULL` for both cycle types.**
- New migration `20260518020000_domain_application_domain_required` backfills `domainId` from `ChallengeVersion.domainId` for existing Standard rows, then tightens the column. Fails fast if any row is missed rather than silently leaving the schema mid-state.
- Standard write paths in `portal.apply.tsx` now set `domainId` alongside `challengeVersionId`. Seed updated (24 sites).
- Single source of truth: "which domain is this DA for" is always `da.domainId`, regardless of cycle type.

**2. Collapsed the `da.challengeVersion!.domainId` non-null bangs.**
- All 10 sites in `portal.apply.tsx` simplified to `da.domainId`. `challengeVersionId` is now purely "questions the applicant answered" — no longer doing double duty as the domain anchor.

**3. Status route + InternToFull notification fan-out are now atomic.**
- `applicationCycleStatusUpdate.create` + `notification.createMany` + `internsNotifiedAt` marker all run inside `prisma.$transaction`. A partial failure can no longer leave the cycle Open with no notifications sent (or notifications sent but `internsNotifiedAt` unset, which would allow re-spam on the next Open).

775/775 tests passing locally. The `?? ??` fallback ladders in `lead.cycle.$id.tsx` / `reviewer.application.$id.tsx` / `release.ts` are now semantically equivalent to just `da.domain` since `domainId` is always populated — left them in place to keep this commit scoped, can clean up in a follow-up.

Kiran-Jones:
Pushed `da08d81` to fix the failed e2e: the earlier seed pass updated every DA using `engCv.id` / `designCv.id` / `pmCv.id`, but missed Dana's DA in the Winter 2028 cycle that uses `engCv2.id` (the v2 challenge version). Added the missing `domainId: engDomain.id` on that one fixture.

Verified by grepping every `challengeVersionId:` line in `prisma/seed.ts` — all 27 DA writes now have a paired `domainId:` line. 775/775 tests still green.

Kiran-Jones:
Pushed two more commits closing the deferred review-pipeline gaps:

**`a9294d0` — Reviewer dashboard: cycle picker**
Reviewers assigned on both Standard and InternToFull cycles now see a small segmented control in the page header to switch between them. `?cycle=<id>` query param drives selection; defaults to first match (Standard preferred). Single-cycle reviewers see no picker — UI is unchanged for them.

**`070f427` — Domain-lead dashboard: InternToFull support**
Two bugs in one:
- 6 prisma queries filtered DAs via `challengeVersion.domainId`, silently excluding InternToFull DAs. Now use `OR [{ challengeVersion: { domainId } }, { domainId }]` via a shared `daDomainMatch` alias.
- `activeCycle` picker now honors `?cycle=<id>`, prefers Standard so existing single-cycle UX stays put, with the same `CyclePicker` component inline next to the domain header.
- `initialDelibsCount` is forced 0 for InternToFull (no Initial round); `finalDelibsCount` uses "all reviews submitted" instead of "completed interview" for InternToFull, matching the delibs route's qualifying filter.

Standard-only sections (challenge versions, interviews) render empty for InternToFull cycles because the underlying queries return empty — functional but a follow-up could trim those empty cards visually.

775/775 tests passing locally. No new tests for the picker UI; existing tests cover the loader-level filter changes via the cycles/delibs paths.

PR #541, opened 2026-05-14
Add RFC 7591 Dynamic Client Registration for MCP clients

Kiran-Jones:
## Why

PR #538 shipped DALI MCP v1 with two pre-seeded `OAuthClient` rows (`claude-desktop`, `claude-code`). When the user actually tried to use them, Claude Code's MCP SDK returned:

> SDK auth failed: Incompatible auth server: does not support dynamic client registration

The SDK has no hardcoded `client_id` for DALI OS and always POSTs to `registration_endpoint` first. The seeded rows were never reachable.

## What

- **`POST /oauth/register`** (RFC 7591). Unauthenticated by spec.
- Accepts the minimal RFC 7591 body. Validates:
  - `redirect_uris` is a non-empty array of `http://127.0.0.1` or `http://localhost` URIs. Rejects `https://`, `0.0.0.0`, public IPs.
  - `token_endpoint_auth_method` defaults to / must be `"none"` (public client).
  - `grant_types` (if present) must be `["authorization_code"]`.
  - `response_types` (if present) must be `["code"]`.
- **Hard-codes the MCP policy server-side** on the created `OAuthClient` row regardless of input:
  - `allowedProviders: ["google"]`
  - `requiredAccountType: "member"`
  - `requireMembership: true`
  - `allowedScopes: ["mcp:read", "mcp:write"]`
  - `isLoopback: true`, `isFirstParty: false`
- Rate-limited to 5 registrations per IP per hour.
- Advertises `registration_endpoint` in `/.well-known/oauth-authorization-server`.
- Drops the dead `claude-desktop` / `claude-code` seeds from `seed.ts` and `seeds/v0-reference.ts`, plus a defensive migration that deletes them from existing DBs **only if** they have no associated `OAuthGrant`.

## Security model

DCR being unauthenticated is fine here because the gates on what a registered client can *do* live on `requireMembership` + `requiredAccountType` at `/oauth/authorize` and `/oauth/callback/google` — and those fields are forced server-side at registration time, not read from the request. The only client-supplied data on the row is `redirect_uris` (validated loopback-only) and `client_name` (displayed on the consent screen, length-capped at 200 chars).

## Test coverage

- Unit (`app/routes/__tests__/oauth.register.test.ts`): happy path; default `client_name`; rejects https loopback, `0.0.0.0`, public IPs, empty/missing `redirect_uris`, `client_secret_basic`, `client_credentials` grant type, `token` response type; rate-limit at 6th request.
- E2E (`e2e/mcp-dynamic-registration.spec.ts`): register → authorize → consent → token → `/mcp` whoami; plus the same negative cases against the live server.

## Migrations

- `20260514160000_delete_dead_mcp_client_seeds` — `DELETE FROM "OAuthClient" WHERE "clientId" IN ('claude-desktop','claude-code') AND NOT EXISTS (...)`. Data-touching but bounded; safe (user confirmed those rows have no grants).


Kiran-Jones:
Fix push: mcp-auth-flow.spec.ts hard-coded the seeded `claude-code` client_id; updated it to register a client via /oauth/register first. Each register call uses a randomized X-Forwarded-For so the 5/hour/IP rate-limit bucket doesn't bleed across tests or retries.

PR #557, opened 2026-05-17
Fix delibs view: live updates, review notes, non-lead access

Kiran-Jones:
## Summary

Fixes four issues in the hiring delibs view:

- **Status doesn't update without refresh** — the domain-lead kanban and the reviewer mirror view both now poll the loader every 5s via `useRevalidator`. The domain-lead kanban resyncs its local `columnOrder` from the latest loader data, skipping the sync while the user is actively dragging or while a move POST is in flight so optimistic state isn't clobbered.
- **Live delibs view for non-lead members not updating** — same 5s `useRevalidator` poll on the reviewer dashboard.
- **Application notes not showing in delib reviews** — render `ApplicationReview.annotations` in `ApplicantContextModal`, grouped by field with the quoted excerpt and the reviewer's comment.
- **Non-domain leads can't read unassigned reviews** — make the reviewer mirror's delibs cards clickable, opening the same `ApplicantContextModal` the domain lead uses. `/full-context` is already gated by `hasCycleAccess` (cycle reviewers pass), so this gives non-leads visibility into every review on apps in their domain.

No schema changes, no migrations, no API contract changes — `annotations` was already on the `/full-context` response (Prisma includes scalars by default), it just wasn't rendered.

Safe to deploy mid-session: concurrent moves reconcile to the server's authoritative `columnOrder` within 5s, and the resync is guarded against in-flight drags/saves.

## Test plan

- [ ] Open delibs as domain lead A in one browser and as domain lead B in another; move a card in A and verify B sees the move within ~5s without refreshing
- [ ] Open delibs as a non-lead reviewer in a third browser; verify their mirror view updates with the same lag
- [ ] Click a card in the non-lead reviewer mirror; verify the `ApplicantContextModal` opens with answers, all reviews, decisions
- [ ] In a review, add a highlight + comment on an answer; open the applicant in delibs and verify the annotation appears under the review as "field label / quoted excerpt / comment"
- [ ] Drag a card while a poll is firing and confirm the drop doesn't get clobbered by a revalidation

### DIFF
--- a/dali-api/app/hiring/components/delibs/ApplicantContextModal.tsx
+++ b/dali-api/app/hiring/components/delibs/ApplicantContextModal.tsx
@@ -124,6 +124,7 @@ export function ApplicantContextModal({
                   ...(data.rubric.generalCriteria ?? []),
                   ...(data.rubric.domainCriteria ?? []),
                 ]}
+                fieldContext={buildFieldContext(data)}
               />
               <DecisionsSection decisions={data.decisions ?? []} />
             </>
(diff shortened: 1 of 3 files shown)


Kiran-Jones:
Duplicated pr

PR #639, opened 2026-05-23
Notify interviewers on assignment so it surfaces in tasks

Kiran-Jones:
## Summary
- Assigned interviews never appeared in the interviewer's home tasks banner or bell because no Notification was created at assignment time. Tasks only surface unread Notifications, so the assignment was invisible until the interview started.
- Fire a Notification to each newly-assigned interviewer at assignment time. It lands in both the bell and the home tasks banner (existing `listOpenTasks` plumbing) and the recipient can dismiss it the usual ways — open the link, mark read, or "mark all read". The interview itself remains the source of truth in `/interviewer/...`.
- No schema change. Uses `NotificationKind = General` with `dueAt = interview.startTime` so the home banner renders the deadline chip.

## Why dismissable, not direct-query
First pass made tasks query `InterviewAssignment` directly so the row clung to the home banner until `endTime` passed. The interviewer flagged that as undesirable — they want a one-time-fires notification they can dismiss after acknowledging, with the interview itself living in the `/interviewer/...` view.

## Implementation
- New helper `app/hiring/lib/interview-notifications.ts` :: `notifyInterviewAssigned({ interviewId, cycleInterviewerIds, createdByUserId? })`.
- Fires **outside** the assignment transaction (`.catch(() => {})`) — mirrors the existing `sendInterviewInviteEmails` / `sendReassignmentEmails` best-effort pattern, so a notify failure can't roll back a committed scheduling change.
- Wired into all 4 assignment-creating paths:
  - Initial schedule — `api.domain-applications.$id.schedule-interview.ts`
  - Reschedule — `api.my-interview.reschedule.ts`
  - Auto-reassign on decline — `reassignInterviewer` in `scheduling.ts`
  - Manual hiring-lead reassign — `api.interviews.$id.reassign.ts`

## Test plan
- [ ] Book an applicant onto a slot → both newly-assigned interviewers see a bell notification + home tasks card linking to `/interviewer/interview/<id>`.
- [ ] Open the link / mark read / "mark all read" → notification clears as expected.
- [ ] Applicant reschedules → fresh notification fires to the new pair.
- [ ] An interviewer declines → the replacement gets a notification (auto-reassign path).
- [ ] Hiring lead manually reassigns → the new interviewer gets a notification.
- [ ] If a notify fails (DB hiccup), the scheduling write still commits and the API still returns 201.

### DIFF
--- a/dali-api/app/hiring/lib/interview-notifications.ts
+++ b/dali-api/app/hiring/lib/interview-notifications.ts
@@ -1,0 +1,90 @@
+import { prisma } from "~/lib/db";
+
+// Emit an in-app Notification to each newly-assigned interviewer so the
+// assignment shows up in the bell + Home tasks banner. Recipients dismiss
+// it like any other notification (open the link, mark read, or "mark all
+// read"); the interview itself stays in their `/interviewer/...` view as
+// the persistent record.
+//
+// Best-effort and runs OUTSIDE the assignment transaction (mirroring
+// sendReassignmentEmails): a flaky write here must not roll back a
+// committed scheduling change. Callers should `.catch(() => {})` it.
+// Safe to call with an empty `cycleInterviewerIds`.
+
+const LOCATION_LABEL: Record<string, string> = {
+  PodAppa: "Pod Appa",
+  PodMomo: "Pod Momo",
+  Online: "Online",
+};
+
+function formatStart(d: Date): string {
+  return d.toLocaleString("en-US", {
+    weekday: "short",
+    month: "short",
+    day: "numeric",
+    hour: "numeric",
+    minute: "2-digit",
+    timeZone: "America/New_York",
+  });
+}
+
+export async function notifyInterviewAssigned(args: {
+  interviewId: string;
+  cycleInterviewerIds: string[];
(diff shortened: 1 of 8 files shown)


Kiran-Jones:
Reworked the approach. The original direct-query design made the task undismissable until the interview's end time, which wasn't right. Now it fires a Notification at assignment time (best-effort, outside the tx) so it surfaces in the bell + home banner via the normal pipeline and clears like any other notification. Wired into all 4 assignment paths: initial schedule, reschedule, auto-reassign-on-decline, manual hiring-lead reassign. `tasks.ts` reverted to its dev state.

Kiran-Jones:
Added the schema link (option 2 from the discussion above). New migration `20260523000000_add_notification_interview_assignment_link` adds nullable `Notification.interviewAssignmentId` with `ON DELETE SET NULL`. `listOpenTasks` / `countOpenTasks` now exclude linked notifications whose assignment is no longer Active or whose interview is no longer Scheduled, so reassign / decline / cancel / complete clear the old interviewer's tile automatically with zero fan-out writes. Unlinked notifications pass through unchanged. 889 tests pass, typecheck clean.

PR #681, opened 2026-05-24
Form drafts + calendar/Manual Blocks follow-up fixes

sophp516:
Follow-up work that landed after #679 was squash-merged. Rebased cleanly onto current `dev` (the commits are cherry-picked, so this contains only net-new changes — no stale reverts).

## Form drafts (new feature + migration)
- `Form` gains `draftQuestions` (JSONB) + `draftIntro` (TEXT): an editable working copy separate from frozen `FormVersion`s. "Save" writes the draft; "Save as version" freezes it into an immutable version and clears the draft. Drafts are **never** served to fillers — published fills always read the latest `FormVersion`.
- Editor wiring in `FormDetail` / `ChallengeBuilder`; supporting changes in `forms-data`, `SlotColumnMapper`, `SubmissionDatabase`, and the intent-to-work / project-bids routes.
- New test: `forms-data.draft.test.ts`.
- **Migration** `20260524120000_add_form_draft_columns` — **additive only** (two nullable columns on `Form`, default NULL). No data loss; existing forms unaffected.

## Calendar / Manual Blocks polish
- Grid lines render **beneath** events (Busy blocks were being cut by the line redraw).
- Manual Blocks: fixed the `+ Cancel` icon (now `✕ Cancel` when open); replaced the raw RRULE text field with the friendly **Repeats** dropdown (Does not repeat / Daily / Weekly / Monthly); fixed Start/End datetime inputs overflowing the form (`min-w-0` + `w-full`).
- Drag-create popover anchors via a callback-ref (a shared `useRef` left the left-flipped anchor null); now-line restyle.

## Library
- Hiring Library section toggle restyled.

## Testing
- `npm run typecheck`: clean (after `prisma generate` — required since this adds schema columns).
- `npm test`: **968/968** pass, incl. the new draft test.
- Migration reviewed: additive, non-destructive.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---
[...]

PR #684, opened 2026-05-24
Harden domain-lead Rubrics e2e against iframe navigation race

sophp516:
Follow-up to #681 — this one-line e2e hardening was pushed just after #681's squash-merge cut, so it didn't make it into `dev`. (#681's other commits, incl. the forms-data Prisma/CI fix, did land.)

`domain-lead.spec.ts` "can reach Rubrics from the sidebar" clicked the Rubrics tab immediately after framing into the Library iframe, without waiting for the route to load inside it — racing the iframe navigation, which made the test hard-fail/flake. Wait for the Rubrics tab to be visible before clicking.

The Library markup is unchanged — the `Rubrics` tab and the `Evaluation Rubrics` heading both still exist — so this is a timing fix, not a regression.

**Caveat:** the e2e suite needs the seeded Postgres CI container, so I couldn't run it locally; this is a reasoned fix verified only to parse/list correctly. No app code changes — e2e spec only.

---
<a href="https://app.blacksmith.sh/dali-lab/codesmith/dali-os/pr/684"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1782232927&installation_id=133523038&pr_number=684&repository=dali-lab%2Fdali-os&return_to=https%3A%2F%2Fgithub.com%2Fdali-lab%2Fdali-os%2Fpull%2F684&signature=c93b5d46237fbb313215c11c4000674982ff78650cfde7abbc5c4e0543865aec"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>
<sup>Need help on this PR? Tag <code>@codesmith</code> with what you need. Autofix is disabled.</sup>

### DIFF
--- a/dali-api/e2e/domain-lead.spec.ts
+++ b/dali-api/e2e/domain-lead.spec.ts
@@ -53,7 +53,12 @@ test.describe('domain lead workflow', () => {
     // tab; Rubrics live behind the in-page Rubrics tab.
     await page.getByRole('button', { name: 'Library' }).click();
     const libraryFrame = page.frameLocator('iframe[title="Library"]');
-    await libraryFrame.getByRole('tab', { name: 'Rubrics' }).click();
+    // Wait for the Library route to finish loading inside the iframe before
+    // interacting — otherwise the Rubrics tab click can race the iframe's
+    // navigation and land on a not-yet-ready frame.
+    const rubricsTab = libraryFrame.getByRole('tab', { name: 'Rubrics' });
+    await expect(rubricsTab).toBeVisible();
+    await rubricsTab.click();
     await expect(libraryFrame.getByRole('heading', { name: 'Evaluation Rubrics' })).toBeVisible();
   });
 });


sophp516:
**My visibility-wait fix did NOT resolve it — the Rubrics test still hard-fails (both attempts).** I was wrong that this is a simple iframe-load race, and I'd rather stop guessing than push a third speculative patch. Here's what I've established and where I'm stuck.

### What the new failure tells us
With `await expect(rubricsTab).toBeVisible()` in place, the test now gets **past** the tab being visible and **past the click** (lines 60–61 pass) — it fails only at the `Evaluation Rubrics` heading (line 62). So: the Rubrics tab renders and is clickable, but the click isn't switching the panel to RubricsPanel.

### What I ruled out
- **Library markup**: unchanged — the `Rubrics` tab (`role=tab`) and the static `<h2>Evaluation Rubrics</h2>` both still exist on `dev`.
- **Tab-switch logic**: byte-for-byte identical between #602 (when this test passed) and now — same `useSearchParams`/`selectTab`/`setSearchParams`, same `{tab === "rubrics" && <RubricsPanel/>}`. My batches only restyled the toggle (classNames).
- **Embed detection** in `layout.tsx` (`searchParams.get('embed')` + `isClientEmbedded`): predates my changes; my diff only touched `isCore`/`photoUrl`.

So nothing in the failing path changed in this PR's lineage.

### Why I think this is environmental, not a regression
Same run: `hiring-lead` "Setup" and `calendar Monday` **flaked** on the identical symptom ("iframe content not visible in 5s"), and the server log is spamming `No routes matched "/hiring/interviewer"` — a route error crashing requests mid-run. Two of those tests are unrelated to anything here. This looks like CI iframe/timing instability under 2-worker load, with the Rubrics test landing on the unlucky side this run.

### Why I can't just confirm it
I **cannot run e2e locally** (needs the seeded Postgres container) and the run's Playwright artifacts (trace/`error-context.md` with the DOM snapshot) are **gone** (`total_count: 0` — expired/not uploaded). So I can't see what the iframe actually contained at failure.

### Options (need a maintainer call)
1. **Re-run the job.** If it goes green, it's flake — merge #684 (the wait is still a reasonable hardening).
2. **Get me the trace.** If someone re-runs with artifact upload (or bumps Playwright retries/timeout), the `error-context.md` ARIA snapshot would settle whether the iframe rendered the shell vs. the bare Library, or whether `?tab=rubrics` took effect.
3. **Quarantine** this + the 2 flaky nav specs (`test.fixme`/retries) as a separate stabilization task, so it stops blocking unrelated PRs.

I don't want to guess a #3 fix into the dark. Happy to pursue any of these with a pointer.

PR #737, opened 2026-06-01
Alumni: Dartmouth directory sync foundation

Kiran-Jones:
## Summary

First PR of the alumni-accounts track (see `alumni_plan.md`). Lays the foundation for deriving alumni status from Dartmouth's People API, with no user-visible behavior change yet — derivation logic, sidebar variants, and the alumni directory ship in #739.

**July 6 rework — People-API-only.** `lookup.dartmouth.edu` went behind Dartmouth SSO (even `GET /` 302s to `saml2/authenticate`), so the original two-API design is dead. Live-record testing (three weeks post-Commencement) showed the People API carries everything in its **no-scope** payload:

| Signal | Behavior |
|---|---|
| `affiliations[]` contains `"Alum"` | Degree conferred — appears within **weeks** of Commencement (the prompt signal) |
| `affiliations[]` contains `"Student"` | **Lingers after graduation** — enrolled-right-now is the compound `Student AND NOT Alum`, which correctly identifies +1 students |
| `dartmouth_affiliation` | Stays `DART` for months post-grad; `ALUMNI` is the long tail |
| `department_class` | `'27`-style class identity for classYear population (not grad year — a '25 on a +1 stays `'25`) |

- **Schema**: 4 nullable cache columns on `User` (`dartmouthAffiliation`, `dartmouthIsAlum`, `dartmouthIsStudent`, `dartmouthPeopleSyncedAt`)
- **Clients**: `dartmouth-jwt.ts` (cached JWT exchange, **no optional scope** — the optional scopes gate FERPA fields we never read and require data-steward approvals), `dartmouth-people.ts` (People API → affiliations + classYear)
- **Orchestrator**: `dartmouth-refresh.ts` — lazy (7-day staleness, matched to the 14-day trust window in #739's derivation), stamps `graduatedAt` on the first observed graduation signal, never clobbers user-set `classYear`, 404 keeps last-known signals
- **Login hooks**: fire-and-forget `refreshDartmouthSignals` from both CAS callbacks **and both Google callbacks** (members mostly sign in with Google; safe no-op without a netId)
- **Env**: `DARTMOUTH_API_KEY` in `.env.example`. Must be set as a Fly secret in staging + prod before merge; when unset the sync silently no-ops and derivation falls back to `graduatedAt` + classYear math.

Migration is one-way `ADD COLUMN` of four nullable fields — no backfill, no defaults, safe under concurrent writes. (The migration was rewritten in place during the rework — it has only ever applied to disposable per-PR preview branches.)

**Known limitation** (documented in `alumni_plan.md`): a BE dual-degree candidate (AB conferred, still enrolled) is indistinguishable from a fresh graduate in the API. #739's derivation protects them via a current-assignment guard.

## Test plan

- [x] `npm test` — 1,402 tests pass (includes 17 unit tests for the JWT exchanger + People client, exercising the four live-record shapes)
- [x] `npm run typecheck` — zero errors in touched files
- [x] Live API verification (2026-07-06): four real netids across current student / +1 term / fresh standard grad / +1-year grad; lookup.dartmouth.edu SSO gate confirmed
- [ ] After merge: set `DARTMOUTH_API_KEY` Fly secret in staging + prod, watch login logs for the fire-and-forget refresh


Kiran-Jones:
Updated (July 6): merged `dev` (conflicts in `.env.example` + `auth.callback.cas.ts` — both were parallel additions, kept both sides). Also dropped the `urn:dartmouth:people:read.sensitive` scope from the JWT exchange: per the People API docs, `dartmouth_affiliation` — the only field we read — is in the base no-scope payload, and requiring the optional scope would break the whole People path if the key isn't granted it (it needs data-steward approvals). `npm test` 1,400 passing; typecheck clean in touched files.

Kiran-Jones:
Reworked (July 6, follow-up): **lookup.dartmouth.edu is SSO-gated** (even GET / 302s to saml2/authenticate), so the lookup client could never have worked from Fly — removed it entirely. The People API's no-scope payload covers everything: `affiliations[]` "Alum" = degree conferred (appears within weeks; IDM code lags by months), "Student" lingers post-grad so enrolled = Student AND NOT Alum, and `department_class` supplies classYear. Verified against four live records. Schema now caches `dartmouthIsAlum`/`dartmouthIsStudent` instead of the two lookup columns (migration rewritten in place — only ever applied to preview branches). Google callbacks now trigger the refresh too. 1,402 tests green.

